PR #5, opened 2020-12-10
Get selected value

errael:
It seems that the two commits are combined in the diffs, can click on a commit to see it separately.

The first commit adds some debug to SSDBComboBox. And introduces setLogColumnName; it was added that after engaging with a wild goose around some `null` in log entries. Also, get rid of `boundColumnType` in SSCommon.

The second  gets rid of settingSelectedItem/selectedItem. And there's a comment about simplifying the "else" in setSelectedItem which I believe can be done.

### DIFF
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/TestBaseComponents.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/TestBaseComponents.java
@@ -317,7 +317,6 @@ public class TestBaseComponents extends JFrame {
 			// SETUP NAVIGATOR QUERY
 				final String query = "SELECT * FROM swingset_base_test_data;";
 				cmbSSDBComboNav = new SSDBComboBox(connection, query, "swingset_base_test_pk", "swingset_base_test_pk");
-				cmbSSDBComboNav.setLogColumnName("combo_nav");
 
 				try {
 					cmbSSDBComboNav.execute();
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/TestBaseComponents.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/TestBaseComponents.java
@@ -185,7 +185,9 @@ public class TestBaseComponents extends JFrame {
 	/**
 	 * combo navigator and sync manger
 	 */
-	SSDBComboBox cmbSSDBComboNav = new SSDBComboBox(); // SSDBComboBox used just for navigation
+	// TODO: this gets set again, why set it here? listeners?
+	//SSDBComboBox cmbSSDBComboNav = new SSDBComboBox(); // SSDBComboBox used just for navigation
+	final SSDBComboBox cmbSSDBComboNav; // SSDBComboBox used just for navigation
 	SSSyncManager syncManager;
 
 	/**
@@ -315,6 +317,7 @@ public class TestBaseComponents extends JFrame {
 			// SETUP NAVIGATOR QUERY
 				final String query = "SELECT * FROM swingset_base_test_data;";
 				cmbSSDBComboNav = new SSDBComboBox(connection, query, "swingset_base_test_pk", "swingset_base_test_pk");
+				cmbSSDBComboNav.setLogColumnName("combo_nav");
 
 				try {
 					cmbSSDBComboNav.execute();
--- a/swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
@@ -81,6 +81,11 @@ import com.nqadmin.swingset.models.SSListItemFormat;
  * represents a foreign key to another table, and the combobox needs to display
  * a list of one (or more) columns from the other table.
  * <p>
+ * <b>Warning. This combobox may use GlazedLists. Do not use methods
+ * that are based on index in the list. Unless you're sure...</b>
+ * setSelectedIndex(-1) is OK, though it may end up at selected
+ * index 0, which is "null/empty" if allowing null.
+ * <p>
  * Note, if changing both a rowSet and column name consider using the bind()
  * method rather than individual setRowSet() and setColumName() calls.
  * <p>
@@ -1347,6 +1352,9 @@ public class SSDBComboBox extends JComboBox<SSListItem> implements SSComponentIn
 				// The next statement either selects
 				// the first item in combo, or a null.
 				super.setSelectedItem(nullItem);
+				if (nullItem == null) {
+					logger.debug(() -> String.format("%s : Setting null when null not allowed. Current editor text is '%s'", getColumnForLog(), getEditor().getItem()));
+				}
 				// if (nullItem != null) {
 				// 	super.setSelectedItem(nullItem);
 				// } else {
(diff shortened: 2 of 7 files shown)


swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -873,11 +878,15 @@ public Long getSelectedValue() {
...
 //				}
 //			}
-			if (settingSelectedItem && selectedItem != null) {
-				result = remodel.getMapping(selectedItem);
-			} else if (getSelectedIndex() != -1) {
-				result = remodel.getMapping(getSelectedIndex());
+			Object item = getSelectedItem();
+			if (item instanceof SSListItem) {
+				result = remodel.getMapping((SSListItem)item);
 			}
```
errael, line 884: I wonder about the evolution of settingSelectedItem/selectedItem. Now that mapping is part of SSListItem maybe that's why it's not needed. Or maybe because null are handled differently.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1331,21 +1341,23 @@ public void setSelectedItem(final Object _value) {
...
+				logger.debug("{}: Selected Index AFTER super.setSelectedItem()={}", () -> getColumnForLog(), () -> getSelectedIndex());
 			} else {
 				if (nullItem != null) {
 					super.setSelectedItem(nullItem);
 				} else {
+					// TODO: why not super.setSelectedItem(null)?
+					//       BTW: can't make this happen now
 					logger.debug("{}: No matching list item found so not updating. Current editor text is '{}'", () -> getColumnForLog(), () -> getEditor().getItem().toString());
 				}
 			}
```
errael, line 1353: I put a breakpoint in the listener, and when entering garbage, getSelectedIndex is -1 and getSelectedItem is null. As noted can't make the log happen, but I have seen it before.

swingset-demo/src/main/java/com/nqadmin/swingset/demo/TestBaseComponents.java
```diff
@@ -315,6 +317,7 @@ public void performRefreshOps() {
 			// SETUP NAVIGATOR QUERY
 				final String query = "SELECT * FROM swingset_base_test_data;";
 				cmbSSDBComboNav = new SSDBComboBox(connection, query, "swingset_base_test_pk", "swingset_base_test_pk");
+				cmbSSDBComboNav.setLogColumnName("combo_nav");
```
errael, line 321: Otherwise there's a `null` in the log entries.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: > and getSelectedItem is null

Wrong, it's index 0 item `{"",null}`. I had tried getSelectedValue() which returns null. 

I'm going to try and find out how that happens after entering a garbage character.

swingset-demo/src/main/java/com/nqadmin/swingset/demo/TestBaseComponents.java
bpangburn: I like it. Maybe ** combo navigator ** so that it stands out from normal db column?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: settingSelectedItem was definitely a hack to deal with glazedlist. If the above works then maybe we can eliminate settingSelectedItem?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1331,21 +1341,25 @@ public void setSelectedItem(final Object _value) {
...
-					logger.debug("{}: No matching list item found so not updating. Current editor text is '{}'", () -> getColumnForLog(), () -> getEditor().getItem().toString());
-				}
+				// Note that nullItem is null when allowNull is false.
+				// The next statement either selects
+				// the first item in combo, or a null.
+				super.setSelectedItem(nullItem);
+				// if (nullItem != null) {
+				// 	super.setSelectedItem(nullItem);
+				// } else {
+				// 	logger.debug("{}: No matching list item found so not updating. Current editor text is '{}'", () -> getColumnForLog(), () -> getEditor().getItem().toString());
```
bpangburn, line 1353: Maybe we can/should still do a logger.trace on the editor text?

bpangburn: Clicked "Start a review" just to see how that was different from a comment.

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
```diff
@@ -82,6 +84,26 @@
  * reverted using Undo button (has to be done manually by the user).
  */
 public class SSDataNavigator extends JPanel {
+	private static class RowSetState {
+		private boolean inserting;
+	}
+
+	// don't have to worry about concurrency, always EDT
+	private static Map<RowSet,RowSetState> rowSetState = new WeakHashMap<>();
```
bpangburn, line 92: Brilliant!

bpangburn: So one map for class SSDataNavigator containing parings of RowSet instances each with the RowSetState for that instance. Each RowSet should only have one navigator and vice versa. 1-1. This allows use of static throughout. If we have 20 screens the map would presumably have 20 entries?

bpangburn (approved):
I'm approving the merge of these changes, but not sure if I'm looking a the draft PR or if there were multiple PRs.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: Put back log with adjusted text

swingset-demo/src/main/java/com/nqadmin/swingset/demo/TestBaseComponents.java
errael: OK, in SSSyncManager constructor I'll set one that stands out if it hasn't already been set. And it can be changed later.

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
errael: Yep, I was assuming 1-1. If there are 20 navigators, should be 20 entries.

I don't think this code cares if there is more than one navigator for a row set, a row set would still only have one state struct. (And if more than one, something else might break first.

Oh, hang on a second, the code pattern is
```
onInsertRow = false;
setInserting(rowSet, onInsertRow);
```
to make it easier to back out. and this pattern means multiple navigators for same row set, would get out of sync with the state. Now that there's by in for this technique, I'll change it to
```
setInserting(rowSet, false);
```
 and get rid of the `onInsertRow` variable.

errael: Considering creating an interface called `RowSetState` which gives read only access to the state for a given rowset, and then ...

I'll leave it as is, but if/when more state is required, can expose RowSetState.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: Thanks.

errael:
Added a commit to handle previous review issues.

bpangburn (approved):

PR #19, opened 2020-12-11
SSComboBox handle allow null.

errael:
Added `setSelectedMapping()` and `getSelectedMapping()`
but did not deprecate `setSelctedValue()` and `getSelectedValue()`.
Fix these methods: get can return null (not NON_SELECTED)
and set handles nullItem/null.

Glazed no longer Strict, but CONTAINS.

Added adustForNullItem. This is used from setOptions and
after setAllowNull. So things are not so sensitive to ordering;
I don't understand the ordering requirements and/or reasoning.
I haven't tested changing setAllowNull when the null is selected
so I probably have to do a bit more.

Big Question:

```
In SSDBComboBox there's
    if (getAllowNull() && (getBoundColumnName()!=null)) {
        // add the nullItem
    }
```
In SSComboBox the check for column name is gone.

Why is it there in SSDBComboBox? Should it be in SSComboBox

Just noticed that setSelectedMapping check to skip work is wrong,
so definitely this PR will be amended.


bpangburn:
> Big Question:
> 
> ```
> In SSDBComboBox there's
>     if (getAllowNull() && (getBoundColumnName()!=null)) {
>         // add the nullItem
>     }
> ```
> 
> In SSComboBox the check for column name is gone.
> 
> Why is it there in SSDBComboBox? Should it be in SSComboBox

This was done so that we never add a null/blank first item for a combo navigator field. There may be a better way to enforce this in SSDBComboBox.

bpangburn (commented):
Do we have to give consideration to get selectedIndex() when artificially adding a null value? I anticipate that some existing code using SwingSet might be calling getSelectedIndex() where mappings are [0 to N-1] as getSelectedIndex() and getSelectedValue() would previously return the same thing unless in the un-selected state (getSelectedIndex()==-1).

I'll check to see how much of that exists.

errael:
> > Big Question:
> > ```
> > In SSDBComboBox there's
> >     if (getAllowNull() && (getBoundColumnName()!=null)) {
> >         // add the nullItem
> >     }
> > ```
> > 
> > 
> > In SSComboBox the check for column name is gone.
> > Why is it there in SSDBComboBox? Should it be in SSComboBox
> 
> This was done so that we never add a null/blank first item for a combo navigator field. There may be a better way to enforce this in SSDBComboBox.

OK. So I don't have to worry about having taken it out for SSComboBox. Good news. I'll focus on SSComboBox for a more dynamic getAllowNull(). Consider SSDBComboBox later.

errael:
> Do we have to give consideration to get selectedIndex() when artificially adding a null value? I anticipate that some existing code using SwingSet might be calling getSelectedIndex() where mappings are [0 to N-1] as getSelectedIndex() and getSelectedValue() would previously return the same thing unless in the un-selected state (getSelectedIndex()==-1).
> 
> I'll check to see how much of that exists.

With nullItem, and glazed list filtering, get/set selectedIndex won't work. (I'm really not too sure about when the glazed lists stuff kicks in, so it might be OK with that (assuming no nullItem) but I wouldn't want to count on it.

getSelectedMapping, aka getSelectedValue, must be used.

So seems like the consideration is that nullItem must be disabled until getSelectedIndex is gone.

bpangburn:
> > I'll check to see how much of that exists.
> 
> With nullItem, and glazed list filtering, get/set selectedIndex won't work. (I'm really not too sure about when the glazed lists stuff kicks in, so it might be OK with that (assuming no nullItem) but I wouldn't want to count on it.
> 
> getSelectedMapping, aka getSelectedValue, must be used.
> 
> So seems like the consideration is that nullItem must be disabled until getSelectedIndex is gone.

So only 289 matches :(
Many are checking for -1 so definitely some work to do there, but in reality, it may only be a few hours of work.

What if we bring in the code, but override getSelectedIndex() to throw an exception in getSelectedIndex if getAllowNull()==true?

That would give me a migration path with a failsafe.

errael:
> What if we bring in the code, but override getSelectedIndex() to throw an exception in getSelectedIndex if getAllowNull()==true?
> 
> That would give me a migration path with a failsafe.

Should work. It looks like getSelectedIndex is only used in JComboBox for "accessible" stuff, so if that's not an issue for you, we should be cool. In other situations it uses model.get...

Could only throw the exception if the return is not -1, but might be better to just bite the bullet.

getSelectedMapping/getSelectedValue return null in two situations. Nothing is selected or the nullItem is selected. We seem to be treating both those the same in general; I don't see why that would be a problem but... `getSelectedItem() == null` means no selection.

bpangburn:
> getSelectedMapping/getSelectedValue return null in two situations. Nothing is selected or the nullItem is selected. We seem to be treating both those the same in general; I don't see why that would be a problem but... `getSelectedItem() == null` means no selection.

That's a useful distinction so probably good to mention this in JavaDoc for SSDBCombo and SSCombo (presuming it holds true for both).

errael:
Closing because I'm having trouble pushing, will recreate branch.

PR #21, opened 2020-12-11
SSComboBox handle allow null. getAllowNull() checks metadata.

errael:
This handles checking the metadata isNullable and use that for getAllowNull. Tested by changing database schema, and see the nullItem in SSComboBox change. Overrode with setAllowNull, that worked. Still more to try... If setAllowNull that stays in effect until explicitly changed. Whereas isNullable changes whenever there's a new rowset or bind.

I decided to cache the result of the metadata check (I stepped through it and it seemed a lot to get metadata). There's might be better ways, can discuss that.

Also, wanted/needed to adjust whether there's new metadata.


swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -366,12 +382,22 @@ public void setModel(ComboBoxModel<SSListItem> model) {
...
 	protected OptionMappingSwingModel<Integer, String, Object> getOptionModel() {
 		return optionSwingModel;
 	}
 
+	@Override
+	public int getSelectedIndex() {
+		//if (ssCommon != null && getAllowNull()) {
+		//	throw new IllegalStateException("getAllowNull() true");
+		//}
+		return super.getSelectedIndex();
```
errael, line 396: This is a little tricky. My initial analysis was flawed, and JCombo uses this a lot. I'll check the stack to see if this is called from "outside". Does nqadmin, other than swingset, identify the app?

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -914,6 +987,57 @@ public void setSelectedValue(final Integer _value) {
...
+	public void bind() {
+		SSComponentInterface.super.bind();
+		adjustForNullItem();
+	}
+
+	@Override
+	public void bind(RowSet _rowSet, String _boundColumnName) {
+		SSComponentInterface.super.bind(_rowSet, _boundColumnName);
+		adjustForNullItem();
+	}
```
errael, line 1038: Need to catch metadata changes. Does this seem reasonable, in my tests bind() didn't get called, but bind(rs, name) did. Is there something else I should catch?

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -405,6 +410,19 @@ protected void bind() {
...
+		//
+		// TODO: is this the right place to do this?
+		//       Will this catch any metadata changes
+		//
+
+		// This is used a lot, just get it now.
+		isNullable = RowSetOps.isNullable(rowSet, boundColumnIndex);
+
+		logger.trace(() -> String.format("Column isNullable: %s.", isNullable));
+
```
errael, line 425: Is this an OK place to catch all metadata changes. Think this covers bind and setRowSet. Anywhere else?

errael:
@bpangburn  I had a few specific things, but if you have time to scan in general...

I think the only thing missing is to add some code in adjustForNullItem to handle if the nullItem is selected and it goes away and vice versa. And fix up the outside getSelectedIndex check.

errael:
I think I should have pulled more context for the comments. Probably easier to just read them in context in the commit diff.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
bpangburn: com.nqadmin is in the package name for jdbcrowsetimpl and for our internal apps. Wondering if GlazedList calls it too?

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
bpangburn: Empty bind() didn't make sense at first, but I guess the idea was to set the pieces (column, rowset) and then bind. I made SSCommon.bind() private and commented it out of SSComponentInterface and no compile errors.

I found it in 3 spots in our app code, but just in comments referring conceptually to bind() so we can probably persist the changes to SSCommon and SSComponentInterface().

SSDataGrid and a few deprecated formatting subpackage components roll their own bind.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
bpangburn: I think SSCommon.bind() is the right spot. setBoundColumnName() and setBoundColumnIndex() call bind() if needed.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -525,6 +546,9 @@ public String getBoundColumnName() {
 	 *
 	 * @return String containing the value in the bound database column
 	 */
+	//
+	// TODO: javadoc seems wrong, looks like allowNull is *TRUE* by default
+	//
 	public String getBoundColumnText() {
```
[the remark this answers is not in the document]
bpangburn: Yes. I originally had it false by default and changed it. Missed this.

bpangburn:
> I think I should have pulled more context for the comments. Probably easier to just read them in context in the commit diff.

Have my IDE and using Ctl-L to jump to the line to keep up :)

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
bpangburn: I had another thought on this, but seems dangerous especially if other stuff is relying on getSelectedIndex(). We could see if mappings are null (meaning 0 to n-1 mappings) **and** getAllowNull **and** if super.getSelectedIndex() >0 then return getSelectedIndex-1. It was a bad idea in my head and worse now that I've actually typed it out. Ugh.

bpangburn: I may just have to suck it up and fix our code to not use getSelectedIndex(). As an intermediate measure we could add a new temporary method such as getSelectedIndexNoFirstNull() that did some adjustment as outline above: if getAllowNull()==true then return -1 for -1 or 0, otherwise return getSelectedIndex()-1.

errael: I've got everything working last night, doing final reviews, may not have time to push until this afternoon (afternoon here). Use of getSelectedIndex throws as discussed.

I'm confused about what you're looking for. I though the issue with getSelectedIndex was for [0-N) mappings. Where you wanted to get the mapping by using the index. So
```
Integer iInt = getSelectedMapping();
int indexIsMapping = iInt != null ? iInt : -1;
```
This doesn't work if it's not a [0,N) mappings.

In any case, feel free to add any transitional methods you need. I wouldn't (and can't) stop you

errael: Perhaps the simplest, would be to put getAllowNull(false) in the constructor, stuff should work as is. Then if you want to use whatever the metadata says, the caller could do combo.setAllowNull(null) which means use the metadata. Or use combo.setAlowNull(true) to put the nullItem in no matter what the metadata says.

errael:
```
boolean noSelection = combo.getSelectedMapping() == null;
```
Is equivelent to
```
boolean noSelection = combo.getSelectedIndex() == -1;
```
in the pre nullItem days.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
errael: > guess the idea was to set the pieces (column, rowset) and then bind

What does this capability contribute? Feels like allowing setting of mappings/options separately; it just opens the door to bugs and maintenance problems for no good reason.

Get rid of it?

bpangburn: Yes please.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
errael: OK. In the future, could generalize and cache more metadata queries, make them lazy. Here's the spot to flush that non-existent cache.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -366,12 +388,34 @@ public void setModel(ComboBoxModel<SSListItem> model) {
...
+				|| caller.startsWith("com.nqadmin.swingset.demo")
+			) && ssCommon != null && ssCommon.getAllowNull()) {
+			throw new IllegalStateException("App::getSelectedIndex && getAllowNull()");
+		}
+
+		return super.getSelectedIndex();
+	}
+
+	
+
```
errael, line 418: Tweak or disable as needed.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -914,6 +1006,106 @@ public void setSelectedValue(final Integer _value) {
...
+	/**
+	 * Catch this to make some adjustments after a change in metadata.
+	 * {@inheritDoc }
+	 */
+	@Override
+	public void metadataChange() {
+		SSComponentInterface.super.metadataChange();
+		adjustForNullItem();
+	}
+
```
errael, line 1108: Use this instead of bind(rs, col) plus setRowSet(rs)

swingset/src/main/java/com/nqadmin/swingset/utils/SSComponentInterface.java
```diff
@@ -139,6 +128,13 @@ default void bind(final RowSet _rowSet, final String _boundColumnName) {// throw
 
 	}
 
+	/**
+	 * Override this method for notification of a change in metadata.
+	 * Typically from setRowSet() or bind().
+	 */
+	default void metadataChange() {
+	}
+
```
errael, line 137: Formalize notification of metadata change. Seems better than inferring that there's a change.

bpangburn (approved):

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
bpangburn: I know we already merged this, but do you recall what this comment was referring to?

errael: I'll adjust the comment so it indicates looking for change in nullability.

PR #23, opened 2020-12-14
Move SSComboBox null handling into abstract SSBaseComboBox.

errael:
Seems to work the same. Wanted to get this out for any testing this week. SSDBComboBox is the next step.


bpangburn (approved):
This really cleans things  up a lot.I was wondering if you were going to have he base class implement SSComponentInterface. Seems like the right call.

PR #24, opened 2020-12-14
Ssdb combo box use base

errael:
Two commits. The first has some SSDBComboBox bug fixes. The second as SSDBComboBox use SSBaseComboBox.


swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -78,6 +78,19 @@
...
+//       - protected class ComboBoxListener ....
+//       - protected final ComboBoxListener ssComboBoxListener = new ComboBoxListener();
+//       - getSelectedMapping()
+//       - public void addSSComponentListener() {...}
+//       - public void removeSSComponentListener() {...}
+//       - public void customInit() {...}
+//       - getMappings()
+//       - getOptions()
+//       - 
+//
```
errael, line 93: Some or all of these is easy.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -268,13 +284,22 @@ public void metadataChange() {
...
+
 	/**
 	 * Add or remove the nullItem from the itemList if needed depending
 	 * on {@code getAllowNull()} and the current existence of a nullItem.
 	 * If logical null is selected before, keep it selected after.
 	 */
 	protected void adjustForNullItem() {
-		boolean wantNull = getAllowNull();
+		boolean wantNull = getAllowNull() && !isComboBoxNavigator();
 		boolean hasNull = nullItem != null;
```
errael, line 303: This is the idea?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1008,6 +951,24 @@ private boolean hasOption2() {
...
+	/**
+	 * After this, make some adjustments.
+	 * {@inheritDoc }
+	 */
+	@Override
+	public void setBoundColumnName(String _boundColumnName) {
+		ssCommon.setBoundColumnName(_boundColumnName);
+		adjustForNullItem();
+	}
+
```
errael, line 971: This is my understanding.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1306,6 +1269,12 @@ public void setSecondDisplayColumnName(final String _secondDisplayColumnName) {
...
 	 * @param _value value to set as currently selected.
 	 */
+	//
+	// TODO: The code currently assumes that _value is an SSListItem.
+	//       But the description above makes me dubious of that claim.
+	//       Is _value a string? Should the type be checked and
+	//       and then proceed accordingly?
+	//
 	@Override
 	public void setSelectedItem(final Object _value) {
```
errael, line 1279: There may be some work needed around here.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1343,7 +1313,11 @@ public void setSelectedItem(final Object _value) {
 		//
 		// 2020-10-03_BP: Updated getSelectedValue() to properly return the primary key rather than using getSelectedIndex() during a call to this method.
 		// Only try to update item for a valid list item.
-			if (selectedItem!=null) {
+
+			// TODO: when this method is called internally,
+			//       newSelectedItem may be nullItem. Should
+			//       the other branch be taken in that case?
+			if (newSelectedItem!=null) {
```
errael, line 1320: ?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1504,31 +1478,52 @@ public void setSelectedItem(final Object _value) {
 	 *
 	 * @param _value value to set as currently selected.
 	 */
+	//
+	// TODO: WHAT IS THIS SUPPOSED TO DO IF BOTH O,O2 are set?
+	//
+	// TODO: thought this was deprecated/removed. Maybe that was SSComboBox
+	// 
 	public void setSelectedStringValue(final String _value) {
```
errael, line 1486: There's some considerations.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1504,31 +1478,52 @@ public void setSelectedItem(final Object _value) {
...
-		// ONLY NEED TO PROCEED IF THERE IS A CHANGE
-		// TODO consider firing a property change
-		try (Model.Remodel remodel = comboInfo.getRemodel()) {
+		// TODO consider firing a property change ??? there are already events
+
+		try (Model.Remodel remodel = optionModel.getRemodel()) {
+			// TODO: If getAllowNull, might be empty, but still have 1 item.
+			// TODO: Why bother checking if empty? If it's empty,
+			//       then there is nothing expensive that happens in here.
 			if(!remodel.isEmpty()) {
```
errael, line 1494: The code works (I think) if it's empty. Why check?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1542,51 +1537,66 @@ public void setSelectedStringValue(final String _value) {
...
 		// 2020-08-03: Removing conditional as this could be called when consecutive records
 		// have the same value and we want to make sure to update the editor Strings
 
 		// IF MAPPINGS ARE SPECIFIED THEN LOCATE THE SEQUENTIAL INDEX AT WHICH THE
 		// SPECIFIED CODE IS STORED
 
-		try (Model.Remodel remodel = comboInfo.getRemodel()) {
+		try (Model.Remodel remodel = optionModel.getRemodel()) {
+			// TODO: if getAllowNull, might be empty, but still have 1 item.
 			if (!remodel.isEmpty()) {
```
errael, line 1563: Need a better check for empty. Maybe `hasValidItems()` which takes nullItem into account.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: If identical, they should all be easy.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: Yes.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: Prior to SwingSet 4, setSelectedItem() was only ever called once and that was with null. I will see if/how it is used in our app. It was not overridden previously and I think it was only overridden here to work around odd behvaior with GlazedLists. Looks like settingSelectedItem() has been commented out so I'm thinking we may not even need to override setSelectedItem() any more??

bpangburn: No, maybe it's sitll needed. I didn't scroll down enough. The selectAll() is useful, although it's possible we could relocate it.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: If the user has selected nullItem then I think the code should proceed as though there has been a valid selection, not null. I think null was expected in the case where the user typed gibberish into a GlazedList combo.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: I don't even see it in SSComboBox 3 or 4. It was in SSDBComboBox 3.x. Will add it to my list of things to check.

bpangburn: In 3, this method was one line of code that set the text in the underlying bound JTextField:
this.textField.setText(_value);

I feel like something has gotten out of hand one this one.

Considering the old days of primary keys (Mappings) and display strings (Options) and you (the developer) want to set the combo based on a target display string (Option), how would you do that? I think we've discussed that it's a dangerous methods because there could be more than one string match (Option) so it may not behave as intended.

I would find/set the first Option that matches or log a warning and setSelectedIndex(-1) or setSelectedItem(null).

If this is not used a lot then maybe we can deprecate.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: Agree. See notes above. This used to be a one line method.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1542,51 +1537,66 @@ public void setSelectedStringValue(final String _value) {
...
+				//setSelectedIndex(index);
+				
 				logger.trace("{}: eventList - [{}].", () -> getColumnForLog(), () ->  remodel.getItemList().toString());
 				logger.trace("{}: options - [{}].", () -> getColumnForLog(), () ->  remodel.getOptions().toString());
 				logger.trace("{}: mappings - [{}].", () -> getColumnForLog(), () ->  remodel.getMappings().toString());
-				
-				setSelectedItem(index == -1 ? null : remodel.get(index));
-				//setSelectedIndex(index);
 			} else {
 				logger.warn(getColumnForLog() + ": No mappings available for current component. No value set by setSelectedValue().");
```
bpangburn, line 1584: Can/should we call setSelectedItem(null) here?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: If you call getMappings() on an empty remodel it will throw an NPE? Or maybe calling indexOf() throws NPE when there are no mappings?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1638,18 +1637,18 @@ public void setSSCommon(final SSCommon _ssCommon) {
...
 	 * @return returns true if update is successful otherwise returns false.
 	 */
 	public boolean updateOption(final Long _mapping, final String _option) {
 
 		boolean result = false;
 
-		try (Model.Remodel remodel = comboInfo.getRemodel()) {
+		try (Model.Remodel remodel = optionModel.getRemodel()) {
 			final int index = remodel.getMappings().indexOf(_mapping);
-			if (index < 0) {
```
bpangburn, line 1652: Looks like the old code was: 
// 				// PROCEED IF INDEX WAS FOUND
// 				if (index != -1) {

bpangburn (commented):
Overall, I think we're in good shape. I need to see if/how  setSelectedItem() and setSelectedStringValue() are being used in our app.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: For some random reason I put in
```
SSListItem newSelectedItem = (SSListItem) _value;
```
Changing this to
```
Object newSelectedItem = _value;
```
And I think everything works ok. Escpecially if don't want to go into `== null` branch.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: Notice the `==null` branch does `setSelectedItem(nullItem)` so nullItem is selected in both branches for either null or nullItem. Only the log entry is different and there's enough log info to know. So just getting rid of the cast should be ok.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: I'm thinking this is the right thing to do, but that depends on the callers intent. It's all about what to do if the item isn't found. Should it just be put into the editable combo box?
```
					Object item;
					if (index != -1) {
						item = remodel.get(index);
					} else {
						//item = nullItem;
						// Didn't find it in the list,
						// so just use it for setSelectedItem.
						item = _value;
						logger.warn(() -> String.format("%s: Could not find a corresponding combobox item for %s. Using it for selectedItem.", getColumnForLog(), _value));
					}
					
					// setSelectedIndex()->setSelectedItem() For glazed reliability.
					// setSelectedIndex(index);

					setSelectedItem(item);
```
So either a listItem that matches something in the itemList is selected or whatever string was passed in is put into the editable combo box according to setSelectedItem rules.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: `getMappings()` on empty model returns an empty list. If the list has has a nullItem, then a list of one element, `null`, is returned. Same with `getOptions()`, except you either empty list or list with one item an empty String.

 Hmm, does that seem right?

That's by examination, not running a test.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: I'm still confused. And what if both O and O2 are set?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: Do you mean in the case when the list is empty? I think null is selected automatically in that case by JComboBox. Wouldn't hurt to select null.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: So I went to `< 0` (a bug) to `>= 0` (just weird).

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: It's not used in our application code at all.

bpangburn: Not used in our app, although I guess someone might use it since we're overriding a public method so it should probably behave.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: We use this 4 times each with a string so presumably not the multi column scenario. If O2 is set then I would throw an exception.

I think this method could be helpful if you have a relatively small list and no O2. Probably add a bit to JavaDoc to indicate it will fail if you configure O2 and could give unexpected results if there are multiple matches. Maybe move it to SSBaseComboBox? It seems it would be more useful for an SSComboBox.

Could deprecate in favor of a setSelectedOption()?

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: Yes. When remodel.isEmpty()==true.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: Was just trying to determine if if '(!remodel.isEmpty()) {' is necessary, but I think it is to respond if there are no mappings. Just to confirm will we have [0-n) mappings if custom mappings are not specified? If so then we're really just worried about giving a log warning if the combo does not have any items (ignoring nullItem), thus your suggestion of hasValidItem().

This can happen in our app in practice.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: I don't think putting it in the combo editor helps, although maybe that give the user some useful feedback?

errael: At this point, unless there's a clear reason, I'm thinking that we give things the feel of a combo box. So if the app does `setSelectedOption("bogus_value")`, and "bogus_value" is not a known option, then do setSelectedItem("bogus_value"). It's on the app.

errael: > then do setSelectedItem("bogus_value"). It's on the app.

When stuff migrates into SSBaseCombo, then it becomes more clear what changes/adjustments are to expected behavior. Then it might make sense to specialize things more.

bpangburn: I'm good with consistency and I'll squawk if it breaks things.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: > will we have [0-n) mappings if custom mappings are not specified?

Yes. Good point. It's highly unlikely for there to be an empty list; bad app.

But we do have the situation where `getMappings()`/`getOptions()` returns a list that and initial null/empty item when `getAllowNull()` is true.

bpangburn: Definitely will have this in practice. We have one where you select a parent record and it filters a SSDBComboBox of children. Sometimes there are no children.

errael:
@bpangburn Added a commit on top of the others. You can look only at that to see cleanup in response to review. 

Will move stuff into SSBaseComboBox next. Can either request change, merge this, or wait for moving stuff into combo box.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1549,19 +1526,17 @@ public void setSelectedStringValue(final String _value) {
 	 */
 	public void setSelectedMapping(final long _value) {
 
-		// HANDLED BY UNDERLYING INFRASTRUCTURE: TODO consider firing a property change
-		// HANDLED: TODO what if user pass null or if getSelectedValue() is null?
-
+		// TODO: IS THIS COMMENT CORRECT? multiple items with same mapping?
 		// 2020-08-03: Removing conditional as this could be called when consecutive records
```
bpangburn, line 1530: I agree. I don't think that 2020-08-03 comment is right. May have been copy/paste or have mixed up value/items/mappings/options. Would not expect duplicate mappings.

bpangburn (approved):
Just one comment agreeing with your comment about a comment. I'll approve/merge and maybe you can fix the JavaDoc with the next commit?

PR #25, opened 2020-12-14
Handling of null and empty string on insert

bpangburn:
@errael 

RowSetOps.updateColumnText() was not handling empty strings well on the insert row for columns that were not null based on metadata.

SSFormattedTextField had its own popups if getAllowNull() was false. Now won't give the warning on the insert row.


errael (approved):
Glad isInserting() is handy(). Being able to get hold of state with out passing stuff around ... That kind of decoupling is the idea behind  CentralLookup.

Change makes sense from my extremely limited point of view.

PR #26, opened 2020-12-16
Move more stuff into SSBaseComboBox.

errael:
`getMappings()`/`getOptins()` return a list that has first element null or "" when getAllowNull is true. Issue?

There's a list of things to consider moving into SSBaseComboBox just before the class declaration. Comment or just add stuff to the list.

logging in SSBaseComboBox uses the subclass logger.

`setSelectedOption(_option)` only used in SSDBComboBox, but seemed appropriate fore Base. Previously (this morning) it returned if empty, so it wouldn't set _option into combobox editor. Now it does it even if it's empty.

`setSelectedMapping()` some minor differences between SSComboBox/SSDBComboBox. One checked `remodel.isEmpty()`, the other didn't. One checked if setting to the same value, the other didn't. SSBaseComboBox checks both optimizations.

`setSelectedMapping()`/`setSelectedOption()` interesting differences. In setSelectedMapping(_mapping) if there is no match for _mapping, null/nullItem is selected in the combobox; in setSelectedOption(_option) if there is no match then it is used in setSelectedItem(_option) which would appear in the editor. Logging difference to consider.

Example3 gets
```
WARN  setSelectedMapping() - [SUPPLIER_ID]: No mapping available for 1 in combobox, setSelectedItem(null)
```
When the list is "almost" empty, it has nullItem. But by the time the component is displayed it's ok. Not sure if there's an issue or not.

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -39,13 +39,19 @@ package com.nqadmin.swingset;
 
 // SSBaseComboBox.java
 
+import java.awt.Dimension;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import javax.swing.ComboBoxModel;
 import javax.swing.JComboBox;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import com.nqadmin.swingset.models.AbstractComboBoxListSwingModel;
 import com.nqadmin.swingset.models.GlazedListsOptionMappingInfo;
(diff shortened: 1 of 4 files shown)


bpangburn:
> `getMappings()`/`getOptins()` return a list that has first element null or "" when getAllowNull is true. Issue?

I don't think so. I'm not sure these would ever be called outside of SwingSet so maybe they could be made protected? I'll check against app tomorrow.

> There's a list of things to consider moving into SSBaseComboBox just before the class declaration. Comment or just add stuff to the list.
> 
> logging in SSBaseComboBox uses the subclass logger.
Should be fine. For me I want the logger to tell me where the code is that is being executed. I see that SSCombo and SSDBCombo still have their own loggers so I think we'll see the appropriate calling class.

> 
> `setSelectedOption(_option)` only used in SSDBComboBox, but seemed appropriate fore Base. Previously (this morning) it returned if empty, so it wouldn't set _option into combobox editor. Now it does it even if it's empty.
Seems OK. Will have to see if there is odd behavior resulting.

> 
> `setSelectedMapping()` some minor differences between SSComboBox/SSDBComboBox. One checked `remodel.isEmpty()`, the other didn't. One checked if setting to the same value, the other didn't. SSBaseComboBox checks both optimizations.
Seems OK.

> 
> `setSelectedMapping()`/`setSelectedOption()` interesting differences. In setSelectedMapping(_mapping) if there is no match for _mapping, null/nullItem is selected in the combobox; in setSelectedOption(_option) if there is no match then it is used in setSelectedItem(_option) which would appear in the editor. Logging difference to consider.
> 
> Example3 gets
> 
> ```
> WARN  setSelectedMapping() - [SUPPLIER_ID]: No mapping available for 1 in combobox, setSelectedItem(null)
> ```
Maybe this should be an INFO? Now that we're allowing setting of null, if the apps sets null (or the user selects the null first item), I don't think we want a WARN. If the user or code tries to make a bad selection then a WARN seems appropriate. Again, I probably have to play around with it in the demo or our app to see if there is something undesirable.

> 
> When the list is "almost" empty, it has nullItem. But by the time the component is displayed it's ok. Not sure if there's an issue or not.
Sorry. I'm not sure I follow you on this one. By "almost" empty do you mean that it just has nullItem or are you talking about effects from GlazedList filtering down results (or something else)?

A big thing for me is to make sure that for any JavaDoc we spell out what is expected with the edge cases for methods. Otherwise it's too easy to forget, only looking at the code from time to time.

Probably need to merge and test to give more meaningful feedback, but I'll let you decide if any of my feedback above makes you want to discuss further or change anything. If you're good with my comments then feel free to pull the trigger.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -189,6 +221,12 @@ protected BaseGlazedModel() {
 		}
 	}
 
+	/**
+	 * The subclass implements this to provide the logger.
+	 * @return logger
+	 */
+	protected abstract Logger logger();
+
```
bpangburn, line 229: So this will use the logger from the subclass? The way I have the logging configured it shows the calling class so I generally prefer to let each class have it's own logger. That way we can tell if we're looking at a call from Base vs Combo vs DBCombo. My preference would be to give SSBaseComboBox it's own logger.

errael: OK

PR #28, opened 2020-12-17
Clean up of Eclipse warnings.

bpangburn:
@errael Please scan though these at your convenience and merge if appropriate. Warnings about warnings. A lot of noise.


swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -76,9 +76,9 @@
  * are the correct techniques.
  * There are a variety of method to change the current combobox item
  * such as
- * {@link #setSelectedMapping(java.lang.Object) setSelectedMapping(Integer)}.
```
errael, line 79: I originally had `{@link \#setSelectedMapping(Integer)}` but got complaints because it's generic and it wanted to put Object in there. So I rewrote it in two pieces so that `setSelectedMapping(Integer)` is what gets displayed. In the other file, SSDBComboBox, you'll see `setSelectedMapping(Long)`

I don't remember if the complaints came from running maven or the IDE. I'm thinking it must be the IDE, sigh. You're IDE gets it right.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -76,9 +76,9 @@
...
+ * {@link #setSelectedMapping(Integer)} / {@link #setSelectedItem(Object)}.
```
errael, line 79: It might not be a good idea to put `setSelectedItem(Object)` here. Users of this class won't have an easy way to create an SSListItem.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -76,9 +76,9 @@
...
  * <p>
```
errael, line 80: The idea here was to point out that to change the selected item, you probably don't want to use `xxxItem()`, but rather the key as in `xxxMapping`. That should be stated more clearly.

swingset/src/main/java/com/nqadmin/swingset/models/OptionMappingSwingModel.java
```diff
@@ -66,7 +66,7 @@
  * @param <O2>  option2 type; if present, supplementary display string
  * 
  * @see AbstractComboBoxListSwingModel for general description of using this class
- * @see <a target="_top" href="https://javadoc.io/doc/com.glazedlists/glazedlists/latest/ca/odell/glazedlists/swing/AutoCompleteSupport.html">GlazedLists AutoCompletion javadoc</a>
+ * @see <a href="https://javadoc.io/doc/com.glazedlists/glazedlists/latest/ca/odell/glazedlists/swing/AutoCompleteSupport.html">GlazedLists AutoCompletion javadoc</a>
```
errael, line 69: Why? Did you try it and see the difference?

swingset-demo/src/main/java/com/nqadmin/swingset/demo/MainClass.java
```diff
@@ -651,6 +654,7 @@ public static void main(final String[] _args) {
 				case 's': userSqlFiles.add(g.getOptarg());  break;
 				case '?':
 				case 'h':
+				default: // TODO: not sure if default is OK here meaning all other characters indicate an error??
 					some_error = true;
```
errael, line 660: It's cool. g.getopt() never returns anything not in its list. But adding default would catch forgetting to add a case for a new option, rather than silently ignoring.

swingset/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -62,7 +62,7 @@ org.eclipse.jdt.core.compiler.problem.missingJavadocTags=warning
 org.eclipse.jdt.core.compiler.problem.missingJavadocTagsMethodTypeParameters=disabled
 org.eclipse.jdt.core.compiler.problem.missingJavadocTagsOverriding=disabled
 org.eclipse.jdt.core.compiler.problem.missingJavadocTagsVisibility=private
-org.eclipse.jdt.core.compiler.problem.missingOverrideAnnotation=ignore
+org.eclipse.jdt.core.compiler.problem.missingOverrideAnnotation=warning
```
errael, line 65: My IDE doesn't care, maybe there's an option to warn. Generating javadoc shows deprecated. So this warning seems pretty useless. It doesn't bother me if its there. I hope this was automated

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
```diff
@@ -470,7 +470,7 @@ public static void updateColumnText(final RowSet _rowSet, final String _updatedV
 			
 		case DATE:
 // TODO Good to get rid of getSQLDate if possible.
-			if (_updatedValue.length() == 10) {
+			if (_updatedValue!=null && _updatedValue.length() == 10) {
```
errael, line 473: The logic, I guess too difficult for the compiler to follow, insures that this is not null (at least its supposed to). I was getting some null warnings, so add the `assert(_updatedValue != null)` to get rid of them. Maybe can take out the assert now.

errael:
I think I went through everything. Looks good, except for that first commit. Those checkers are more anal than I am.

swingset/src/test/java/com/nqadmin/swingset/models/AbstractComboBoxListSwingModelTest.java
```diff
@@ -261,7 +261,7 @@ public void testCreateListItem() {
 		assertFalse(result.equals(result2));
 	}
 
-	private boolean getElemEquals(Object[] elems, int listItemIndex, LI.Remodel remodel) {
+	private static boolean getElemEquals(Object[] elems, int listItemIndex, LI.Remodel remodel) {
```
errael, line 264: This comment is generally about going to static. I guess it's totally safe when the methods are private. I have a long standing nervousness about using static in some of these cases. I guess it stems from the inheritance issues. But I like the added info you get from declaring static, it references nothing in the instance. Kind of like how I feel about not doing `int i = someVal` if the assignment is never used.

swingset/src/test/java/com/nqadmin/swingset/models/SSListItemFormatTest.java
```diff
@@ -195,32 +195,32 @@ public void testSetPattern() {
...
+		fmt1.setPattern(JDBCType.TIMESTAMP, "timestamp");
 
-		currentPattern = fmt.getPattern(JDBCType.DATE);
+		currentPattern = fmt1.getPattern(JDBCType.DATE);
 		assertEquals(currentPattern, "date");
-		currentPattern = fmt.getPattern(JDBCType.TIME);
+		currentPattern = fmt1.getPattern(JDBCType.TIME);
 		assertEquals(currentPattern, "time");
-		currentPattern = fmt.getPattern(JDBCType.TIMESTAMP);
+		currentPattern = fmt1.getPattern(JDBCType.TIMESTAMP);
```
errael, line 223: fmt --> fmt1 looks good. I don't like hiding instance variables. The "_" helps there. (different file...)

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
bpangburn: Agree. I didn't scroll up enough to see the assertion. I suppressed the warning and added a TODO.

swingset/src/main/java/com/nqadmin/swingset/models/OptionMappingSwingModel.java
bpangburn: Few issues because I can't view internal and external link in the same window at work. However I see one is trying to bring up the external pages in the right frame vs clearing the window. Found that target= is fine if it's after the href.

However, while researching, if target=_blank then it will try a new tab which is nicer in my opinion so giving that a try.

bpangburn: target="_blank" rel="noopener noreferrer" to prevent 'tabnapping'

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -71,14 +71,19 @@
...
+ * There are a variety of methods to change the current combobox item
+ * such as:
+ * {@link SSBaseComboBox#setSelectedMapping(java.lang.Object) setSelectedMapping(Integer)}
+ * {@link SSBaseComboBox#setSelectedOption(java.lang.Object) setSelectedOption(String)}
  * <p>
- * Notice that {@link #getSelectedMapping() }/{@link #getSelectedValue() }
+ * Notice that:
+ * {@link #setSelectedValue(Integer)} and 
+ * {@link SSBaseComboBox#setSelectedMapping(java.lang.Object) setSelectedMapping(Integer)}
  * return null in two situations related to {@link #getAllowNull() }
```
errael, line 87: That would **get**, not *set*.

bpangburn: Operator error. Fix enroute.

swingset/src/main/java/com/nqadmin/swingset/models/OptionMappingSwingModel.java
errael: Cool. I hadn't heard of tabnapping.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
errael: Fixed it before I was done reviewing. What confidence.

PR #33, opened 2020-12-20
Add SSBaseComboBox#hasSelection(), tweak javadoc

errael:


errael:
New method, not used; we're going in.

PR #44, opened 2020-12-28
Issues with text field on data navigator.

bpangburn:
Enter key field navigation into nav editor reverted prior text field's
changes. Buttons not updating properly after entering record number.
added function to eliminate duplicate code to flush changes to database.

### DIFF
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
@@ -119,7 +119,6 @@ public class Example4UsingHelper extends SSFormViewScreenHelper {
 		
 		// Display Screen
 		showUp(_container);
-
 	}
 
 	@Override
@@ -145,10 +144,6 @@ public class Example4UsingHelper extends SSFormViewScreenHelper {
 		txtPartWeight.setPreferredSize(MainClass.ssDim);
 		txtPartCity.setPreferredSize(MainClass.ssDim);
 		
-	// SET COMBO OPTIONS
-		//cmbPartColor.setOptions(new String[] { "Red", "Green", "Blue" });
-		cmbPartColor.setOptions(Arrays.asList(new String[] { "Red", "Green", "Blue" }));
-
 	// SETUP THE CONTAINER AND LAYOUT THE COMPONENTS
 		final Container contentPane = getContentPane();
 		contentPane.setLayout(new GridBagLayout());
@@ -202,6 +197,10 @@ public class Example4UsingHelper extends SSFormViewScreenHelper {
 
 	@Override
 	protected void bindComponents() throws Exception {
+		// SET COMBO OPTIONS
+		//cmbPartColor.setOptions(new String[] { "Red", "Green", "Blue" });
+		cmbPartColor.setOptions(Arrays.asList(new String[] { "Red", "Green", "Blue" }));
+		
 		//txtPartID.bind(getRowset(), "part_id");
 		txtPartName.bind(getRowset(), "part_name");
 		cmbPartColor.bind(getRowset(), "color_code");
--- a/swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
@@ -377,18 +377,23 @@ public class SSDataNavigator extends JPanel {
 			@Override
 			public void actionPerformed(final ActionEvent ae) {
 				logger.debug("FIRST button clicked.");
-				removeRowsetListener();
+				
 				try {
-					if (modification) {
-						if (!dBNav.allowUpdate()) {
-							// UPDATE NOT ALLOWED SO DO NOTHING.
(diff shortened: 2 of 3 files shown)


swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
```diff
@@ -202,6 +197,10 @@ protected void addNewRecordToCmbNavigator() {
 
 	@Override
 	protected void bindComponents() throws Exception {
+		// SET COMBO OPTIONS
+		//cmbPartColor.setOptions(new String[] { "Red", "Green", "Blue" });
+		cmbPartColor.setOptions(Arrays.asList(new String[] { "Red", "Green", "Blue" }));
+
```
errael, line 203: Curiosity. Does moving this make a difference or avoid an issue?

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
```diff
@@ -778,20 +814,30 @@ public void actionPerformed(final ActionEvent ae) {
...
-				removeRowsetListener();
+			public void keyPressed(final KeyEvent ke) {
+				
 				if (ke.getKeyCode() == KeyEvent.VK_ENTER) {
+					
 					try {
+						removeRowsetListener();
+						
+						if (!commitChanges()) return;
+
```
errael, line 825: I see the keyReleased to keyPressed. And this commitChanges().

I know I include minor/cosmetic changes with bug fixes together in the same commit. I shall endeavor to do that no more; it makes it tricky to get to the meat of it.

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
```diff
@@ -848,6 +894,27 @@ protected void addToolTips() {
...
 
 	} // end protected void addToolTips() {
+	
+	/**
+	 * Common code to commit changes to the database from the rowset if
+	 * modifications are allowed.
+	 * @return true unless dBNav.allowUpdate() returns false
+	 * @throws SQLException SQL Exception if rowset call to updateRow() fails
+	 */
+	private boolean commitChanges() throws SQLException {
```
errael, line 904: Cool. It's nice to have that dup'd code gone. I guess this is the guts of the "auto commit"

(I usually just hit add acomment, on the last comment I pushed "start review" to try something new, and got a yellowish button that said pending. I guess that makes sense. Get rid of that soon.

swingset/src/main/java/com/nqadmin/swingset/utils/SSComponentInterface.java
```diff
@@ -145,7 +146,9 @@ default void configureTraversalKeys() {
 		final Set<AWTKeyStroke> forwardKeys = ((JComponent) this)
 				.getFocusTraversalKeys(KeyboardFocusManager.FORWARD_TRAVERSAL_KEYS);
 		final Set<AWTKeyStroke> newForwardKeys = new HashSet<>(forwardKeys);
-		newForwardKeys.add(KeyStroke.getKeyStroke(KeyEvent.VK_ENTER, 0));
+		if (!(this instanceof JTextArea)) {
+			newForwardKeys.add(KeyStroke.getKeyStroke(KeyEvent.VK_ENTER, 0));
+		}
```
errael, line 151: It's been a very long time since I've looked at keyboard focus traversal. So enter was causing problems only with the JTextComponent?

I thought changing keyRelease to keyPressed took care of the issue. Ah, so this change means that enter when in a textcomponent, goes nowhere?

errael (approved):

swingset/src/main/java/com/nqadmin/swingset/utils/SSComponentInterface.java
bpangburn: Not any text component, just JTextArea, which are the multiline text boxes. Otherwise you can use Enter to add line breaks.

swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
bpangburn: Avoids an issue. In my helper bindComponents is called before addComponents. This results in the first record showing SSComboBoxes as blank. If you navigate next/back then it refreshes.

I may add an explicit abstract method like populateSSCombos() to the form helper to guide the developer and call that before bindComponents() or addComponents().

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
```diff
@@ -363,6 +400,52 @@ public SSDataNavigator(final RowSet _rowSet, final Dimension _buttonSize) {
...
+//		}
+//		eventBus.register(new BusReceiver());
+	}
+
+	// TODO: also have Set<SSComponentInterface> modifiedComponents
+	private final Set<SSComponentInterface> errorComponents = new HashSet<>();
+	class BusReceiver {
+// TODO: ENABLE FOR EVENTBUS		
+//		@Subscribe
+//		public void handleRowDataChanged(RowSetModificationEvent ev) {
```
errael, line 423: Not that it matters, but I think you could just have just commented out the `@Subscribe`

Issue #38, opened 2020-12-23
Disable Combobox listeners when the combo popup is visible


bpangburn:
Per Issue #37, we want to disable listeners on the component (which trigger rowset updates, navigation) when the combo popup is visible. This would stop unnecessary rowset updates (bound combo) and record traversals (combo navigator) when the user is typing in the editor or using the arrow keys to traverse the combo items.

Likely includes changes to at least SSBaseComboBox and SSSyncManager. We want to add a popup menu listener that disables some of the other listeners when the popup becomes  visible and re-enable when the popup becomes invisible. When the popup becomes invisible we likely also want to propagate the mapping for the selected item to the rowset (bound combo) or initiate a record navigation (combo navigator).

https://docs.oracle.com/javase/8/docs/api/javax/swing/event/PopupMenuListener.html

bpangburn:
Popup menu and focus listeners were a nightmare. https://stackoverflow.com/questions/10293135/jcombobox-focuslost-is-not-firing-why-is-that

Abandoning.

Issue #37, opened 2020-12-23
Combobox / GlazedList 'contains' anomaly


bpangburn:
When SSComboBox or SSDBComboBox are using GlazedLists (default) there is an issue if the user types a character into the editor that does not match the first character of an item in the list. GlazedLists support is configured for substring matching (.setFilterMode(TextMatcherEditor.CONTAINS)), but there appears to be an anomaly in GlazedLists such that the first character is always checked using a "starts with" strategy.

Per @errael "It uses an array of strategies with only one element, a "startsWithMatcher". Add a second strategy, "contains", it first would try startswith, if that fails it would try contains."

The result is that if a combo contains two items {switch, knob} there will be matches if the user types "s" or "k", but not, for example, "w". If the user types "w", <backspace>, "w" the contains strategy kicks in.

This triggers unnecessary calls to setSelectedItem(null), which also attempt to update the rowset. If a component is configured with setAllowNull(false) then popup warnings appear.

Options include:
1. submitting a patch to GlazedList to modify how the matching strategies are applied
2. adding an enhancement to GlazedLists to support the STRICT and CONTAINS options together (see: https://github.com/glazedlists/glazedlists/issues/676)
3. returning with no action inside setSelectedItem() when getAllowNull()==false AND getModel().getSize() > 0 (ignoring any component update when there are no items in the list)
4. disabling listeners on the component (which trigger rowset updates, navigation) when the combo popup is visible

4 is likely a desired enhancement regardless of this Issue as it would stop unnecessary rowset updates (bound combo) and record traversals (combo navigator) when the user is typing in the editor or using the arrow keys to traverse the combo items.

errael:
There's a draft PR, https://github.com/glazedlists/glazedlists/pull/694, to support strict+contains.

errael:
To verify/check if glazed strict+contains fixes the issues, steps to reproduce the problems are required. This is particularly important considering #45 which is supposed to mask the problems.

errael:
In https://github.com/bpangburn/swingset/pull/45/files#r550332940 there's

> Example 3. Type "am", into Part and we get this in the logs:

Compare to when strict+contains to understand issue.

errael:
I hadn't noticed that this was closed, reopening. I'm still in the middle of working options 1 and 2 from the original report, https://github.com/bpangburn/swingset/issues/37#issue-774024966

In addition, this could be an umbrella issue that includes #60

errael:
[...]

Issue #17, opened 2020-12-11
Mimimize String casting between components and RowSet


bpangburn:
Many SwingSet components use RowSetOps.getColumnText() and updateColumnText() which cast back and forth from java.sql.JDBCType (or java.sql.Types) to String. It would be preferable to eliminate unnecessary castings by storing component values in the appropriate Object type, using RowSet.updateObject() and RowSet.getObject().

errael:
`getBoundColumnText()` is used all over, this in turn uses `getColumnText()`

bpangburn:
> `getBoundColumnText()` is used all over, this in turn uses `getColumnText()`

Not a trivial undertaking, but I think getObject() and updateObject() would be better for non-text components. I can be talked out of this and things are much cleaner than they were before with some consolidation in SSCommon and the cleanup done in SSRowSet. If we're doing something like storing a number in a text field then the casting is fine, but there are other places where it seems unnecessary to me.

PR #45, opened 2020-12-29
More predictable behavior for combos

bpangburn:
Completely off track from the original branch intent (Issue #38). Was hoping to switch from action/item listeners to focus listeners, but as cited here: https://stackoverflow.com/questions/10293135/jcombobox-focuslost-is-not-firing-why-is-that
 the focus listener is very difficult to deal with for combos. This lead me to abandon focus listeners and try to fix the action listeners to better deal with:
1. Glazed List issue #37 
2. "garbage" strings entered by user
3. clearing of the combo by the user (e.g., backspace to clear any text in the combo)
I spent a lot of time trying to auto-revert strings if a CONTAINS matched on a GlazedList resulted in zero items (e.g. "ScreX"). 

At one time (months ago?) the combination of:
```
getEditor().setItem(priorEditorText);
showPopup();
updateUI();
```
seemed to work, but I could not get it to take. The behavior is what I would expect if isEditable()==false, but GL supposedly forces combos to be editable, and I confirmed that isEditable() returned true inside of setSelectedItem().

### DIFF
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example3.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example3.java
@@ -186,7 +186,8 @@ public class Example3 extends JFrame {
 					rs.close();
 
 				// SET OTHER DEFAULTS
-//					 cmbSupplierName.setSelectedValue(0);
+					 logger.debug("Setting default for Supplier Name mapping to 2.");
+					 cmbSupplierName.setSelectedMapping((long) 2);
 //					 cmbPartName.setSelectedValue(0);
 //					 txtQuantity.setText("0");
 
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -131,7 +131,6 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 			// This is just for keeping a bound column in sync.
 			//
 			// Could be combined with next block, but keeping them separate for debugging.
-			// May be able to eliminate actionListenerNoUpdate when GlazedLists fully supports STRICT/CONTAINS
 			if (isComboBoxNavigator()) {
 				logger.debug("{}: Action Listener returning. No bound column.", () -> getColumnForLog());
 				return;
@@ -163,7 +162,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 			// IF SELECTED ITEM IS NOT AN SSLISTITEM, CHECK THE glGlitchItem
 			// 
 			// This seems to be a strange timing issue. The logs indicate that setSelectedItem() below has encountered 
-			// SCENARIO 1-B or 1-C and has made a call to super.setSelectedItem() to select the correct/updated SSListItem, but
+			// SCENARIO 1-B and has made a call to super.setSelectedItem() to select the correct/updated SSListItem, but
 			// selectedItem here still has a String. glGlitchItem has the correct/updated SSListItem.
 			//
 			// Making a 2nd call to super.setSelectedItem(glGlitchItem) appears to resolve the issue and does not trigger an
@@ -312,13 +311,14 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	protected final SSBaseComboBoxListener ssBaseComboBoxListener = new SSBaseComboBoxListener();
 
 	/**
-	 * List item just used only to workaround GlazedList CONTAINS bug.
+	 * List item used to workaround GlazedList CONTAINS bug/glitch.
 	 */
 	// **** GL STRICT/CONTAINS ****
 	private SSListItem glGlitchItem = null;
 	
 	/**
-	 * Indicates that a garbage string has been entered and the Action Listener should not 
+	 * Indicates that a garbage string has been entered and the Action Listener should not
+	 * try to update the rowset.
 	 */
 	// **** GL STRICT/CONTAINS ****
 	private boolean actionListenerNoUpdate = false;
@@ -404,9 +404,8 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	
 	/**
 	 * {@inheritDoc }
-	 * This method override additionally selects a currently edited item in the combo box.
-	 * 
-	 * If there is a string in the combo editor that does not have any matches, it will revert the string.
+	 * Deal with edge cases for combo editor interaction with GlazedLists and lack of support
+	 * for STRICT/CONTAINS in GlazedLists.
 	 */
 	// **** GL STRICT/CONTAINS ****
 	//
@@ -466,12 +465,12 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	        	priorEditorText = currentEditorText;
 	        	logger.debug("{}: Null not from UI (e.g., rowset, default). Selected item after super.setSelectedItem()={}", () -> getColumnForLog(), () -> getSelectedItem());
 	        	
-	        } else if (getModel().getSize() > 0) {
-	        // SCENARIO #1-B ABOVE - null due to GlazedList glitch when items contain the first character, but GL does not match the first item returned
-	        	glGlitchItem = getItemAt(0);
-	        	super.setSelectedItem(glGlitchItem);
-	        	priorEditorText = getEditor().getItem().toString();
-	        	logger.debug("{}: Null due to GlazedLists glitch so selecting first item in model. Selected item after super.setSelectedItem()={}", () -> getColumnForLog(), () -> getSelectedItem());
+			} else if (getModel().getSize() > 0) {
+			// SCENARIO #1-B ABOVE - null due to GlazedList glitch when items contain the first character, but GL does not match the first item returned
+				glGlitchItem = getItemAt(0);
+				super.setSelectedItem(glGlitchItem);
+				priorEditorText = getEditor().getItem().toString();
+				logger.debug("{}: Null due to GlazedLists glitch so selecting first item in model. Selected item after super.setSelectedItem()={}", () -> getColumnForLog(), () -> getSelectedItem());
 
 	        } else if (!currentEditorText.isEmpty()) {
 	        // SCENARIO #1-C ABOVE - null because user likely entered garbage - revert editor and return without a call to setSelectedItem()
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -114,9 +114,6 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	/**
 	 * Listener(s) for the component's value used to propagate changes back to the rowset in certain instances.
 	 * <p>
-	 * The focus listener is the primary driver of rowset updates from the combo box to minimize noise while
-	 * traversing and to gracefully handle nulls.
-	 * <p>
 	 * NOTE: No guaranty whether ActionListener or FocusLister will be called first when focus is lost.
 	 */
 	protected class SSBaseComboBoxListener implements ActionListener, Serializable
@@ -129,37 +126,41 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 		 * {@inheritDoc} */
 		@Override
 		public void actionPerformed(final ActionEvent ae) {
-			
-			// If the ActionListener has to revert a change it will fire the ActionListener again
-			//
+
 			// If this is a combo navigator, SSSyncManager will have it's own listeners.
 			// This is just for keeping a bound column in sync.
-			if (revertingItem || (getBoundColumnName() == null)) return;
-			
-			Object currentItem = getSelectedItem();
-			
-			if (currentItem instanceof SSListItem) {
-				// PROPAGATE LATEST VALUE TO ROWSET
-					updateRowset();
-										
-				// Update revertItem for next use
-					revertItem = (SSListItem) getSelectedItem();
-				
-			} else {
-				// A PROPER ITEM HAS NOT BEEN SELECTED. WARN USER AND REVERT.
-
-				// NOTIFY THE USER
-					JOptionPane.showMessageDialog(SSBaseComboBox.this,
-						"An item was not selected from the combobox. Reverting to " + revertItem);
-				
-				// Revert the selectedItem. This will trigger the action listener so set a flag.
-					revertingItem = true;
-					setSelectedItem(revertItem);
-					revertingItem = false;
-					
-				// There is nothing to do with the rowset - it should already have the last value
+			//
+			// Could be combined with next block, but keeping them separate for debugging.
+			// May be able to eliminate actionListenerNoUpdate when GlazedLists fully supports STRICT/CONTAINS
+			if (getBoundColumnName() == null) {
+				logger.debug("{}: Action Listener returning. No bound column.", () -> getColumnForLog());
+				return;
 			}
-
+			
+			// Could be combined with prior block, but keeping them separate for debugging.
+			// May be able to eliminate actionListenerNoUpdate when GlazedLists fully supports STRICT/CONTAINS
+			if (actionListenerNoUpdate) {
+				logger.debug("{}: Action Listener returning. actionListenerNoUpdate set to TRUE.", () -> getColumnForLog());
+				return;
+			}
+			
+			// EXTRACT SELECTED ITEM
+			Object selectedItem = getSelectedItem();
+			logger.debug("{}: Selected item from inside Action Listener: {}.", () -> getColumnForLog(), () -> selectedItem);
+			
+			// CHECK FOR A STRING - EXPECT EMPTY STRING WHEN THE USER CLEARS THE EDITOR (E.G., BACKSPACE)
+			if (selectedItem instanceof String) {
+				if (!((String)selectedItem).isEmpty()) {
+					logger.warn("{}: Selected item is a String of '{}'. This is unexpected unless the user clears the editor.",
+							() -> getColumnForLog(), () -> selectedItem);
+				}
+				setSelectedItem(glGlitchItem); // GENERALLY SELECTED ITEM IS EMPTY STRING AND glGlitchItem IS FIRST ITEM IN LIST
+				return; // GLAZEDLIST GLITCH - KEEP LAST VALUE AND DON'T UPDATE ROWSET
+			}
+			
+			// UPDATE ROWSET
+			logger.debug("{}: About to update RowSet with {}.", () -> getColumnForLog(), () -> getSelectedItem());
+			updateRowset();
 		}
 	}
 
@@ -240,6 +241,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 			model.autoComplete.setFilterMode(TextMatcherEditor.CONTAINS);
 			// RESTORE JCOMBOBOX UP/DOWN ARROW HANDLING OVERRIDING GLAZEDLIST
 			_jc.glazedListArrowHandler();
+
 			//model.autoComplete.setStrict(true);
 
 			return model;
@@ -288,14 +290,14 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	protected final SSBaseComboBoxListener ssBaseComboBoxListener = new SSBaseComboBoxListener();
 
 	/**
-	 * SSListItem to stash value when the combo gains the focus so it can be reverted if necessary.
+	 * List item just used only to workaround GlazedList CONTAINS bug.
 	 */
-	protected SSListItem revertItem;
+	private SSListItem glGlitchItem = null;
 	
 	/**
-	 * Boolean flag to indicate that the focus listener is reverting the selected item
+	 * Indicates that a garbage string has been entered and the Action Listener should not 
 	 */
-	private boolean revertingItem = false;
+	private boolean actionListenerNoUpdate = false;
 
 	/**
 	 * The combo model.
@@ -408,6 +410,16 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 		
 		// INITIALIZATION
 		Object newSelectedItem = _value;
+		glGlitchItem = null; // WORKAROUND FOR GL 'CONTAINS' GLITCH
+		actionListenerNoUpdate = false; // TRUE WHEN THE USER HAS ENTERED SOMETHING RESULTING IN NO MATCHES
+		
+		logger.debug("{}: setSelectedItem called with {}.", () -> getColumnForLog(), () -> newSelectedItem);
+
+// TODO: This is likely an empty string where the user cleared the editor. Could add a popup.
+//		if (newSelectedItem instanceof String) {
+//			logger.warn("{}: setSelectedItem called with a String of {}. Likely GL glitch. Ignore.", () -> getColumnForLog(), () -> newSelectedItem);
+//		}
+		
 		final String currentEditorText;
 		
 		// GET CURRENT EDITOR TEXT IF AVAILABLE
@@ -426,14 +438,15 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 		// SCENARIO #1 ABOVE - a valid item that should match something in the list (including nullItem, if not null)
 			super.setSelectedItem(_value);
 			priorEditorText = currentEditorText;
-			// TODO See if we can get rid of this conditional to do selectAll() completely.
-//			if (getEditor() != null && hasFocus()) {
-//				// after we find a match, do a select all on the editor so
-//				// if the user starts typing again it won't be appended
-//				// 2020-12-21_BP: if we don't limited to field with focus, the comboboxes blink on navigation
-//				// this also causes the focus to jump out of a navigation combo
-//				getEditor().selectAll();
-//			}
+			// 2020-12-29_BP: CONFIRMED THIS IS NEEDED FOLLOWING A NORMAL ITEM SELECTION
+			//   OTHERWISE USER IS APPENDING EXISTING ITEM STRING IF THEY START TO TYPE
+			if (getEditor() != null && getEditor().getEditorComponent().hasFocus()) {
+				// after we find a match, do a select all on the editor so
+				// if the user starts typing again it won't be appended
+				// 2020-12-21_BP: if we don't limited to field with focus, the comboboxes blink on navigation
+				// this also causes the focus to jump out of a navigation combo
+				getEditor().selectAll();
+			}
 			logger.debug("{}: Valid match. Selected item after super.setSelectedItem()={}", () -> getColumnForLog(), () -> getSelectedItem());
         } else if (!getEditor().getEditorComponent().hasFocus()) {
         // SCENARIO #3-A ABOVE - setting null from a rowset listener or programmatically (e.g., a default) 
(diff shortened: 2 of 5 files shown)


swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -831,33 +870,38 @@ public void setSelectedValue(final Integer _value) {
...
+			final String boundColumnText = getBoundColumnText();
 
-				setSelectedMapping(comboCode);
+			// LOGGING
+			logger.debug("{}: getBoundColumnText() - " + boundColumnText, () -> getColumnForLog());
 
-			} else {
-				//setSelectedIndex(-1);
+			// GET THE BOUND VALUE STORED IN THE ROWSET - may throw a NumberFormatException
+			Integer targetValue = null;
```
bpangburn, line 888: I think the only difference in updateSSComponent() for SSComboBox and SSDBComboBox is that targetValue is Integer in SSComboBox and Long in SSDBComboBox.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -831,33 +870,38 @@ public void setSelectedValue(final Integer _value) {
...
+			if ((boundColumnText != null) && !boundColumnText.isEmpty()) {
+				targetValue = Integer.parseInt(boundColumnText);
```
bpangburn, line 890: vs Long.parseLong()

bpangburn: Could probably have an abstract method of (M)getTargetValue() in SSBaseComboBox and only keep that piece in the implementations, moving the rest of this method into the base class. A conditional seemed messy. Probably a better way.

errael: Is this an aspect of #17? I'd consider splitting RowSetOps.updateColumnText, something like
```
Object RowSetOps.parseText(JDBCTYpe type, String text)

RowSetOps.updateColumnText(...)
    _rowSet.updateObject(parseColumnText(...))
```
And then `(M)RowSetOps.parseText(...)` could work (possible with a check warning).

Adding an abstractMethod like `(M)parseString` is certainly a workaround.

swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example3.java
```diff
@@ -186,7 +186,8 @@ public void performPreInsertOps() {
 					rs.close();
 
 				// SET OTHER DEFAULTS
-//					 cmbSupplierName.setSelectedValue(0);
+					 logger.debug("Setting default for Supplier Name mapping to 2.");
+					 cmbSupplierName.setSelectedMapping((long) 2);
```
bpangburn, line 190: This was just to have a default of something other than null to make sure it was being picked up properly for an insert.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,33 +112,59 @@
...
 			}
-
-			addRowSetListener();
+			
+			// Could be combined with prior block, but keeping them separate for debugging.
+			// May be able to eliminate actionListenerNoUpdate when GlazedLists fully supports STRICT/CONTAINS
+			if (actionListenerNoUpdate) {
+				logger.debug("{}: Action Listener returning. actionListenerNoUpdate set to TRUE.", () -> getColumnForLog());
+				return;
+			}
```
bpangburn, line 145: actionListenerNoUpdate is likely STRICT/CONTAINS  workaround specific. Presumably if GL fully supported it, we'd only ever have null or an SSListItem passed to setSelectedItem(). Will have to put some thought into when we'd expect null (insert row, some escaping from/clearing of the combo editor)? Maybe we'd only ever have null for the insert row OR getModel().getSize()==0 ?

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,33 +112,59 @@
...
+			
+			// EXTRACT SELECTED ITEM
+			Object selectedItem = getSelectedItem();
+			logger.debug("{}: Selected item from inside Action Listener: {}.", () -> getColumnForLog(), () -> selectedItem);
+			
+			// CHECK FOR A STRING - EXPECT EMPTY STRING WHEN THE USER CLEARS THE EDITOR (E.G., BACKSPACE)
+			if (selectedItem instanceof String) {
```
bpangburn, line 152: If GL STRICT/CONTAINS is fully supported, could we ever get a String here?

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,30 +112,73 @@
...
+			// **** GL STRICT/CONTAINS ****
+			//
+			// IF SELECTED ITEM IS NOT AN SSLISTITEM, CHECK THE glGlitchItem
+			// 
+			// This seems to be a strange timing issue. The logs indicate that setSelectedItem() below has encountered 
+			// SCENARIO 1-B or 1-C and has made a call to super.setSelectedItem() to select the correct/updated SSListItem, but
+			// selectedItem here still has a String. glGlitchItem has the correct/updated SSListItem.
+			//
+			// Making a 2nd call to super.setSelectedItem(glGlitchItem) appears to resolve the issue and does not trigger an
+			// additional ActionListener event.
```
bpangburn, line 170: This gave me fits, but  apparently the Action Listener gets out of sync with setSelectedItem(). It makes me wonder if things aren't being delivered to the logger in order.

Example 3. Type "am",<Enter> into Part and we get this in the logs:
15:30:11.070 [AWT-EventQueue-0] DEBUG com.nqadmin.swingset.SSBaseComboBox.setSelectedItem() - [PART_ID]: Null due to GlazedLists glitch so selecting first item in model. Selected item after super.setSelectedItem()={Cam,5}
15:30:11.362 [AWT-EventQueue-0] DEBUG com.nqadmin.swingset.SSBaseComboBox.actionPerformed() - [PART_ID]: ACTION LISTENER: getSelectedItem() has 'am'.
15:30:11.366 [AWT-EventQueue-0] DEBUG com.nqadmin.swingset.SSBaseComboBox.actionPerformed() - [PART_ID]:  -- About to call super.setSelectedItem({Cam,5}) {}.
15:30:11.371 [AWT-EventQueue-0] DEBUG com.nqadmin.swingset.SSBaseComboBox.actionPerformed() - [PART_ID]:  -- getSelectedItem() now returns {Cam,5}.
15:30:11.373 [AWT-EventQueue-0] DEBUG com.nqadmin.swingset.SSBaseComboBox.actionPerformed() - [PART_ID]: About to update RowSet with {Cam,5}.
15:30:11.374 [AWT-EventQueue-0] DEBUG com.nqadmin.swingset.datasources.RowSetOps.updateColumnText() - [PART_ID]. Update to: 5. Allow null? [false]
15:30:11.378 [AWT-EventQueue-0] DEBUG com.nqadmin.swingset.utils.SSComponentInterface.setBoundColumnText() - [PART_ID]: 5
15:30:11.379 [AWT-EventQueue-0] DEBUG com.nqadmin.swingset.SSBaseComboBox.updateRowset() - [PART_ID]: Setting column to 5.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -254,6 +311,18 @@ protected BaseGlazedModel() {
...
+	/**
+	 * List item just used only to workaround GlazedList CONTAINS bug.
+	 */
+	// **** GL STRICT/CONTAINS ****
+	private SSListItem glGlitchItem = null;
+	
+	/**
+	 * Indicates that a garbage string has been entered and the Action Listener should not 
+	 */
+	// **** GL STRICT/CONTAINS ****
```
bpangburn, line 323: As you can see, I'm trying to flag the workarounds with '**** GL STRICT/CONTAINS ****'

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
 		Object item = getSelectedItem();
 		return item != null && item != nullItem;
 	}
+
```
bpangburn, line 404: I rewrote/reordered things here and reworked the comments. I think the description aligns with the code. Note that I have not encountered SCENARIO 1-D or SCENARIO 3 in testing, but felt safer providing some code to handle. The user would get a popup if encountered so we should hear about it.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -563,6 +769,24 @@ public void setAllowNull(boolean _allowNull) {
 		adjustForNullItem();
 	}
 
+// TODO: Review this. GlazedLists is probably still calling it so need to detect those calls.
```
bpangburn, line 772: I added then commented this out. Pretty sure GL choked on it. Probably have to add some code to ignore calls from GL, but would be useful in the future if someone is using a non GL combo and attempts setEditable(true).

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -642,6 +866,26 @@ protected void adjustForNullItem() {
 			}
 		}
 	}
+
```
bpangburn, line 869: Felt like this was better/safer outside of the Action Listener.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: See Issue #46

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -822,6 +822,45 @@ public void setSelectedValue(final Integer _value) {
...
+//	 * Updates the value stored and displayed in the SwingSet component based on
+//	 * getBoundColumnText()
+//	 * <p>
+//	 * Call to this method should be coming from SSCommon and should already have
+//	 * the Component listener removed
+//	 */
+//	@Override
+//	public void updateSSComponent() {
+//		
+//		// TODO Modify this class similar to updateSSComponent() in SSFormattedTextField and only allow JDBC types that convert to Long or Integer
```
bpangburn, line 835: Discussed a bit in SSComboBox and SSDBComboBox. Goal is to move most/all of the logic here, but didn't want to spend time on Integer vs Long vs M so left very similar logic in the child classes. Can rectify later. See Issue #46.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
bpangburn: Opened Issue #46 . This can wait. And it is somewhat related to Issue #17

swingset/src/main/java/com/nqadmin/swingset/utils/SSSyncManager.java
```diff
@@ -72,37 +73,57 @@
...
 				comboPK = comboBox.getSelectedMapping();
+				logger.debug("COMBO NAVIGATOR: getSelectedMapping() returned: {}.", () -> comboPK);
 				
-				// getSelectedMapping() could return null during initialization or 
-				// If using UP/DOWN arrows with a GlazedList and you above first item or below last item. Note that this should
-				// not happen now that glazedListArrowHandler() has been added to SSBaseComboBox.
+				// getSelectedMapping() could return null during initialization.
+				// We check for null/empty rowset in the prior block.
 				if (comboPK==null) {
+					// **** GL STRICT/CONTAINS ****
```
bpangburn, line 105: Handling a little different than the SSBaseComboBox Action Listener since glGlitchItem is not visible here. Seems to be working well. Definitely getting null from substring matching in the logs so it's needed for now.

swingset/src/main/java/com/nqadmin/swingset/utils/SSSyncManager.java
```diff
@@ -72,37 +73,57 @@
...
+				// **** GL STRICT/CONTAINS ****
+				//
+				// Hopefully selectedItem and lastValidItem can be eliminated once GlazedLists fully supports STRICT/CONTAINS
+					// EXTRACT AND STORE SELECTED ITEM
+					Object selectedItem = comboBox.getSelectedItem();
+						
+					// THIS SHOULD ALWAYS BE A SSLISTITEM, BUT COULD BE SOME EDGE CASES?
+					if (selectedItem instanceof SSListItem) {
+						lastValidItem = (SSListItem)selectedItem;
+					} else {
```
bpangburn, line 124: This conditional could probably be replaced with just
`lastValidItem = (SSListItem)selectedItem;`
but playing it safe.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,30 +112,73 @@
...
-				logger.debug(String.format("%s: Setting to %s.",  getColumnForLog(), mapping));
+			// If this is a combo navigator, SSSyncManager will have it's own listeners.
+			// This is just for keeping a bound column in sync.
+			//
+			// Could be combined with next block, but keeping them separate for debugging.
+			// May be able to eliminate actionListenerNoUpdate when GlazedLists fully supports STRICT/CONTAINS
+			if (getBoundColumnName() == null) {
+				logger.debug("{}: Action Listener returning. No bound column.", () -> getColumnForLog());
+				return;
 			}
```
errael, line 138: How about
```
if (isComboBoxNavigator()) {
```

bpangburn: Agree on isComboBoxNavigator(). Pushed an update. A navigator will have it's own listener with possible calls to setSelectedItem(). I think it's best to cleanly exit the method for a navigator as I'm not sure if the two listeners might compete. It's possible this method could mostly be reduced to updateRowset() once STRICT/CONTAINS is resolved so we won't have the Action Listeners making additional calls to setSelectedItem(). Feel safer with the conditional return for now.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -831,33 +870,38 @@ public void setSelectedValue(final Integer _value) {
...
+			// UPDATE COMPONENT
+			if (targetValue==null) {
 				setSelectedItem(nullItem);
+			} else {
+				setSelectedMapping(targetValue);
 			}
 
 		} catch (final NumberFormatException nfe) {
-			logger.warn(getColumnForLog() + ": Number Format Exception.", nfe);
+			logger.error(getColumnForLog() + ": Number Format Exception.", nfe);
```
errael, line 904: I guess it's "impossible" to get here, so no reason to do `setSelectedSomething`

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
```diff
@@ -831,33 +870,38 @@ public void setSelectedValue(final Integer _value) {
...
+			
+			// LOGGING
+			logger.debug("{}: targetValue - " + targetValue, () -> getColumnForLog());
+			
+			// UPDATE COMPONENT
+			if (targetValue==null) {
 				setSelectedItem(nullItem);
+			} else {
+				setSelectedMapping(targetValue);
 			}
```
errael, line 901: Why not just `setSelectedMapping(targetValue)`?

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,30 +112,73 @@
...
-				setBoundColumnText(null);
-				logger.debug(() -> String.format("%s: Setting to null.", getColumnForLog()));
-			} else {
-				setBoundColumnText(String.valueOf(mapping));
-				logger.debug(String.format("%s: Setting to %s.",  getColumnForLog(), mapping));
+			// If this is a combo navigator, SSSyncManager will have it's own listeners.
+			// This is just for keeping a bound column in sync.
+			//
+			// Could be combined with next block, but keeping them separate for debugging.
+			// May be able to eliminate actionListenerNoUpdate when GlazedLists fully supports STRICT/CONTAINS
```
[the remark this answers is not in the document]
bpangburn: It's a duplicate/remnant from the next block. Originally the blocks were together. I'll delete.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,30 +112,73 @@
...
-
-			addRowSetListener();
+			
+			// **** GL STRICT/CONTAINS ****
+			//
+			// Could be combined with prior block, but keeping them separate for debugging.
+			// May be able to eliminate actionListenerNoUpdate when GlazedLists fully supports STRICT/CONTAINS
+			//
+			// actionListenerNoUpdate will be set to true in setSelectedItem() when the user enters a non-matching string.
+			// It is also in the 'else' for unforeseen outcomes. The user is notified with a popup in both cases.
```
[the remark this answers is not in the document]
bpangburn: I couldn't get it to revert the string. I had this working at one point and couldn't figure out why it won't work now, but maybe some interference from GL? After hours on it, I abandoned reverting the string. Now it gives a popup to stop the user from typing and goes to the last GL match.

I'd like to keep the comment here as it provides some insight about what is triggering actionListenerNoUpdate.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,30 +112,73 @@
...
+			Object selectedItem = getSelectedItem();
+			logger.debug("{}: ACTION LISTENER: getSelectedItem() has '{}'.", () -> getColumnForLog(), () -> selectedItem);
+			
+			// **** GL STRICT/CONTAINS ****
+			//
+			// IF SELECTED ITEM IS NOT AN SSLISTITEM, CHECK THE glGlitchItem
+			// 
+			// This seems to be a strange timing issue. The logs indicate that setSelectedItem() below has encountered 
+			// SCENARIO 1-B or 1-C and has made a call to super.setSelectedItem() to select the correct/updated SSListItem, but
+			// selectedItem here still has a String. glGlitchItem has the correct/updated SSListItem.
```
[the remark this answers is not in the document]
bpangburn: I wasted enough time on it. No need for you to. Good news is that the 2nd call to setSelectedItem() seems to resolve it and hopefully at some point in 2021 it will be a moot point and most/all of these workarounds can be deleted.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
...
+	 * 
+	 * If there is a string in the combo editor that does not have any matches, it will revert the string.
+	 */
+	// **** GL STRICT/CONTAINS ****
+	//
+	// IT IS POSSIBLE THAT NO OVERRIDE OF setSelectedItem() WILL BE NEEDED IF GLAZEDLISTS FULLY SUPPORTS
+	// STRICT/CONTAINS
+	@Override
+	public void setSelectedItem(final Object _value) {
+		// 2020-12-30_BP: setSelectedItem outcomes:
```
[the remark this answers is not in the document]
bpangburn: Probably not. I actually have an override of setEditable() to force it to false and block the developer, but it doesn't play well with GL so commented out for now. I'm not sure how this would work with Options/Mappings, but we can explore further by email/Issue you like.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
...
+		if (_value==null) {
+		// #1 - WE HAVE A NULL
+			if (!getEditor().getEditorComponent().hasFocus()) {
+	        // SCENARIO #1-A ABOVE - setting null from a rowset listener or programmatically (e.g., a default) 
+	        	super.setSelectedItem(null);
+	        	priorEditorText = currentEditorText;
+	        	logger.debug("{}: Null not from UI (e.g., rowset, default). Selected item after super.setSelectedItem()={}", () -> getColumnForLog(), () -> getSelectedItem());
+	        	
+	        } else if (getModel().getSize() > 0) {
+	        // SCENARIO #1-B ABOVE - null due to GlazedList glitch when items contain the first character, but GL does not match the first item returned
```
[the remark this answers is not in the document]
bpangburn: I think that's a GitHub/Eclipse thing. It looks fine in my editor. I removed and re-added the tabs and will submit with my commit.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,30 +112,73 @@
...
+			// EXTRACT SELECTED ITEM
+			Object selectedItem = getSelectedItem();
+			logger.debug("{}: ACTION LISTENER: getSelectedItem() has '{}'.", () -> getColumnForLog(), () -> selectedItem);
+			
+			// **** GL STRICT/CONTAINS ****
+			//
+			// IF SELECTED ITEM IS NOT AN SSLISTITEM, CHECK THE glGlitchItem
+			// 
+			// This seems to be a strange timing issue. The logs indicate that setSelectedItem() below has encountered 
+			// SCENARIO 1-B or 1-C and has made a call to super.setSelectedItem() to select the correct/updated SSListItem, but
```
bpangburn, line 166: Just SCENARIO 1-B, not 1-C. Will update.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
 		Object item = getSelectedItem();
 		return item != null && item != nullItem;
 	}
+	
+	/**
+	 * {@inheritDoc }
+	 * This method override additionally selects a currently edited item in the combo box.
```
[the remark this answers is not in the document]
bpangburn: Yes. Gibberish. Maybe a copy/paste? Rewriting.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
...
+	        	logger.debug(() -> String.format("%s: User entered string of '%s' did not match any list items. Attempting to revert to '%s'.",
+	        			getColumnForLog(), currentEditorText, priorEditorText));
+	        
+	        	// TELL ACTION LISTENER NOT TO UPDATE ITEM OR ROWSET
+	        	actionListenerNoUpdate = true; 
+	        	
+	        	// REVERT STRING
+	        	getEditor().setItem(priorEditorText);   
+	        	//showPopup();// When this was working as intended, the ordering of showPopup() before updateUI() was relevant.
+	        	updateUI(); // NEEDED TO SHOW REVERTED ITEM (ORIGINALLY INTENDED TO SHOW REVERTED TEXT IN EDITOR)
```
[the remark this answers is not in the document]
bpangburn: Are you referring to getEditor().setItem(priorEditorText); ?

There is no "getEditor().getEditorComponent().setText()" (at least it's not public).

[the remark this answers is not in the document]
bpangburn: And the selected item at this point in the code is null. getEditor().setItem(priorEditorText) reverts the item (it doesn't behave as expected). It acts more like setSelectedItem(priorSelectedItem). However we're not explicity tracking priorSelectedItem at this point (made up variable name) so I can't make that call. What is should do is just revert the users keystroke, but that doesn't work.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
...
+	        	
+	        	// WARN USER AND LOG
+	        	String editorText = getEditor().getItem().toString();
+	        	
+	        	JOptionPane.showMessageDialog(SSBaseComboBox.this,
+						"The text entered does not match any item in the list. Reverting to '" + editorText + "'.");
```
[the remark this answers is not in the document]
bpangburn: Hopefully this entire method will be greatly reduced or eliminated if we go to STRICT+CONTAINS.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
...
+	        	// TELL ACTION LISTENER NOT TO UPDATE ITEM OR ROWSET
+	        	actionListenerNoUpdate = true; // TELL ACTION LISTENER NOT TO UPDATE ITEM OR ROWSET
+	        	
+	        	// WARN USER AND LOG
+	        	JOptionPane.showMessageDialog(SSBaseComboBox.this, String.format(
+						"Unexpected call to setSelectedItem() for column %s with value '%s'.", getColumnForLog(), _value));
+	        	logger.warn(() -> String.format("%s: Unexpected call to setSelectedItem() with '%s'.)",
+	        			getColumnForLog(), _value));
+	        }
+		} else if (_value instanceof SSListItem) {
```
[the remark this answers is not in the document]
bpangburn: Again, I don't see it in my IDE. Maybe a tab/space thing? Github seems to use 8 character tabs.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
...
+		// #2 - WE HAVE A SSLISTITEM (including nullItem if getAllowNull()==true)
+			super.setSelectedItem(_value);
+			priorEditorText = currentEditorText;
+			// 2020-12-29_BP: CONFIRMED selectAll() IS NEEDED FOLLOWING A NORMAL ITEM SELECTION
+			//   OTHERWISE USER IS APPENDING EXISTING ITEM STRING IF THEY START TO TYPE
+			if (getEditor() != null && getEditor().getEditorComponent().hasFocus()) {
```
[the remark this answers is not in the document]
bpangburn: We do? If the use just does a mouse selection and doesn't type anything in the editor, this is the code that should be executed.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,144 @@ public boolean hasSelection() {
...
+        	actionListenerNoUpdate = true; // TELL ACTION LISTENER NOT TO UPDATE ITEM OR ROWSET
+        	
+        	// WARN USER AND LOG
+        	JOptionPane.showMessageDialog(SSBaseComboBox.this, String.format(
+					"Unexpected call to setSelectedItem() for column %s with value '%s'.", getColumnForLog(), _value));
+        	logger.warn(() -> String.format("%s: Unexpected call to setSelectedItem() with '%s'.)",
+        			getColumnForLog(), _value));
+			
+		}
+	}
```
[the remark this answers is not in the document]
bpangburn: I agree. I don't see how it should land in this else. Hopefully it doesn't, but based on other strangeness, I thought it was good to have a catch all and bring it to the user's attention (and hopefully to our attention).

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: I'm still trying to follow the Eclipse alphabetical approach from there code cleanup tools. Inner classes, declarations (alphabetical), methods (alphabetical). I realize you're not and for good reason in some places. I guess most IDE tools have some sort of field/method explorer.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
bpangburn: Maybe that would be OK. If getAllowNull()==false, does setSelectedMapping(null) work OK?

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
bpangburn: You'd only get here if the database/rowset had non-numeric text. Could be worth a popup?

bpangburn: Adding a popup for SSComboBox and SSDBComboBox.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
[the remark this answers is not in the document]
errael: Using beep rather than dialog discussed elsewhere.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
errael: > does setSelectedMapping(null) work OK?

It seems like it after a quick look. But it's hardly an important issue.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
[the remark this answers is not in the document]
bpangburn: We could try that, but at the moment it's acting consistently so I don't want to spend too much more time on it. When it was reverting text, the updateUI was needed to reflect the removed character. I think if you remove it now, the item is blank until you navigate or do something else to trigger a refresh. This is one of those things that might even vary a bit based on implementation, platform, window manager, etc.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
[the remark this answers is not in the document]
bpangburn: There are 2 of roughly 6 cases where we want this set to true. Seems cleaner to me to initialize to false and then set to true in 2 cases where it's needed rather than introducing break/return in 4 places to prevent it from being set to false. Let's discuss further by email if needed. Maybe there is a best practice I'm missing/convention I'm missing/violating.

swingset/src/main/java/com/nqadmin/swingset/SSComboBox.java
bpangburn: Changed it and it seems to work fined. Committed/pushed updates to SSComboBox and SSDBComboBox.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,30 +112,72 @@
...
+	 * NOTE: No guaranty whether ActionListener or FocusLister will be called first when focus is lost.
 	 */
 	protected class SSBaseComboBoxListener implements ActionListener, Serializable
 	{
 		private static final long serialVersionUID = -3131533966245488092L;
 
-		/** {@inheritDoc} */
+		/**
+		 * For JComboBox ActionListener and ItemListener are very similar.
+		 *
```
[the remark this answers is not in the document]
bpangburn: No. I was (poorly) documenting a decision to go with ActionListener. I added some JavaDoc to clarify. They do similar things, but ActionListener seems to be the more appropriate choice. Either could probably be made to work.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -111,30 +112,72 @@
...
 
 	/**
-	 * Listener(s) for the component's value used to propagate changes back to bound
-	 * database column.
+	 * Listener(s) for the component's value used to propagate changes back to the rowset in certain instances.
+	 * <p>
+	 * NOTE: No guaranty whether ActionListener or FocusLister will be called first when focus is lost.
 	 */
 	protected class SSBaseComboBoxListener implements ActionListener, Serializable
 	{
```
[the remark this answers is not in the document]
bpangburn: Didn't spend too much time searching, but it appears we're only uninstalling/installing inside of SSCommon.updateSSComponent(). The initial value for a component comes from the rowset so this makes sense. We only need to listen for changes after a component gets it's initial value.

I added Issue #54 to investigate further.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,143 @@ public boolean hasSelection() {
...
+		//     -> set actionListenerNoUpdate to true so the Action Listener does not try to update the rowset
+		//     -> warn user and return
+	
+		// INITIALIZATION
+		glGlitchItem = null; // WORKAROUND FOR GL 'CONTAINS' GLITCH
+		actionListenerNoUpdate = false; // TRUE WHEN THE USER HAS ENTERED SOMETHING RESULTING IN NO MATCHES
+
+		// GET CURRENT EDITOR TEXT IF AVAILABLE
+		final String currentEditorText;
+		if (getEditor().getItem() == null) {
```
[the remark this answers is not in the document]
bpangburn: I think I see it now. If getEditor() returned null then `getEditor().getItem() == null` above should throw an NPE. Since it hasn't in testing, I guess we're safe.

I removed and committed.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -332,6 +401,143 @@ public boolean hasSelection() {
...
+	        	logger.debug("{}: Null not from UI (e.g., rowset, default). Selected item after super.setSelectedItem()={}", () -> getColumnForLog(), () -> getSelectedItem());
+	        	
+			} else if (getModel().getSize() > 0) {
+			// SCENARIO #1-B ABOVE - null due to GlazedList glitch when items contain the first character, but GL does not match the first item returned
+				glGlitchItem = getItemAt(0);
+				super.setSelectedItem(glGlitchItem);
+				priorEditorText = getEditor().getItem().toString();
+				logger.debug("{}: Null due to GlazedLists glitch so selecting first item in model. Selected item after super.setSelectedItem()={}", () -> getColumnForLog(), () -> getSelectedItem());
+
+	        } else if (!currentEditorText.isEmpty()) {
```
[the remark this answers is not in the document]
bpangburn: I'll turn on the space/tab display. Some testing indicates that 4 is the default for Eclipse which is odd because digging a bit indicates that tabs should be set to 8 for Java. Explains why I haven't noticed before as I've been using Eclipse for at least a decade.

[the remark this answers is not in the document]
bpangburn: I committed a change to clean up two of the methods, but now thinking this may break the GitHub comments.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: It was at first, but I adapted and I like the consistency.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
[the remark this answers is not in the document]
bpangburn: I was rolling this around in my head some more. You can ignore the above comment about setting to true for 2 of the 6 cases. I didn't get that you were just talking about clean up on exiting the method. It probably could be set to false upon leaving setSelectedItem(), but I don't think it has to be. Will get reset to false upon the next call to setSelectedItem() and the Action Listener shouldn't do anything in the interim.

Issue #52, opened 2020-12-31
Have a NO-LOCKING option for OptionMappingSwingModel


errael:
This applies in particular, when this class is used as a JComboBox/JList model with a huge list. Memory profiling show a huge number of Remodel objects, more that ListItem object (makes sense need remodel for both read and write).

If no-locking, then only need a single remodel object which can be re-used.

Can also consider optimizations for method used directly by JComboBox/JList.

errael:
Need to investigate further, but here's my current understanding. When used with glazed lists, the JComboBox model is using a copy of the item list, so there is no locking concern. If an external thread modifies the EventList it does not immediately affect the JComboBox.

When OptionMappingSwingModel is used directly as a combox model, there is only one list; currently the list is locked when the JComboBox touches it, this is a problem. Here's some options
1. If not used with a glazed list, no locking. Use/re-use a single remodel. Methods used directly by the JCombo box, as defined by the ComboBoxModel, would not need to use a remodel at all. May check EventQueue access only.
2. To enable non-glazed locking, use a copy of the list. Only need to update the copy if the original is changed.
3. Use a delegate DefaultComboBoxModel to install into JComboBox/JList. Not sure what this would get you, but the separation might help.

(1.) above can be done right away with some judicious verification perhaps involving (3.).

errael:
Possibilities 1 and 3 in previous post are implemented, and PR #61 is pending.

In AbstractComboBoxListSwingModel there's a proxy class that can be installed directly into a JComboBox/JList; this proxy delegates to the enclosing class and currently does no locking.

https://github.com/errael/swingset/blob/552e2179e1cf4865dfe814806fa336bf96f9596a/swingset/src/main/java/com/nqadmin/swingset/models/AbstractComboBoxListSwingModel.java#L453

If/when a locking strategy, other than glazedlists, is decided upon this deliniation makes it clear what needs to be addressed. Note that locking is only needed to support multithreaded access to the list data being managed.

Note that these comments only apply to when glazedlists is not used.

Here are some options for locking without using glazed and there are certainly more:
- Use EventQueue.invokeXxx(() -> { try(remodel) {}}) to modify the list. This is probably the simplest. The remodel object doesn't do any locking, it provides a convenient context to make the modificaitons. This could be done now and no changes need to be made to AbstractComboBoxListSwingModel.
- Keep a copy of the list; use conventional locking to manage acess. Update the list used by AbstractComboBoxListSwingModel from the EventQueue
[...]

PR #61, opened 2021-01-07
By default, no OptionMappingSwingModel list locking, re-use Remodel.

errael:
Glazed still handles locking.
https://github.com/bpangburn/swingset/issues/52.
OptionMappingSwingModel as a model created too many Remodel objects.
Move methods take/release lock from Remodel into AbstractModel.
Create/install model proxy; New test file for proxy.

All tests pass, new test created for proxy. Run demo with/without glazed.


bpangburn (approved):
Looks good. Might as well roll into 4 since I'm still awaiting some user feedback.

Issue #62, opened 2021-01-08
Combos throw NPE when rowset has null


bpangburn:

bpangburn:
If the user selects an item from a combo and the existing value was null it throws an NPE:
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at com.nqadmin.swingset.SSBaseComboBox.updateRowset(SSBaseComboBox.java:916)

`if (!getBoundColumnText().equals(tStringMapping)) {`
will throw NPE if getBoundColumnText() is null.

Will push a fix.

PR #63, opened 2021-01-08
Issue #62

bpangburn:
Fix for Issue #62 to deal with null values in getBoundColumnText() before attempting .equals()

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -913,12 +913,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 			//       setBoundColumnText or RowSetOps.updateColumnText
 
 			String tStringMapping = String.valueOf(mapping);
-			String rowsetValue = getBoundColumnText();
-			if (rowsetValue==null) {
-				if (tStringMapping!=null) {
-					setBoundColumnText(tStringMapping);
-				}
-			} else if (!rowsetValue.equals(tStringMapping)) {
+			if (!Objects.equals(getBoundColumnText(), tStringMapping)) {
 				setBoundColumnText(tStringMapping);
 			}
 		}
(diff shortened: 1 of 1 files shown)


swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -913,7 +913,12 @@ private void updateRowset() {
...
-			if (!getBoundColumnText().equals(tStringMapping)) {
+			String rowsetValue = getBoundColumnText();
+			if (rowsetValue==null) {
+				if (tStringMapping!=null) {
+					setBoundColumnText(tStringMapping);
+				}
+			} else if (!rowsetValue.equals(tStringMapping)) {
 				setBoundColumnText(tStringMapping);
 			}
 		}
```
errael, line 924: Looks like
```
if (!Objects.equals(getBoundColumnText(), tStringMapping)) {
```
would work. (but these nested ifs always get to me, so verify...)

errael (changes requested):

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: That looks like it would be cleaner. Wasn't familiar with Objects.equal().

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -913,7 +913,7 @@ private void updateRowset() {
 			//       setBoundColumnText or RowSetOps.updateColumnText
 
 			String tStringMapping = String.valueOf(mapping);
-			if (!getBoundColumnText().equals(tStringMapping)) {
+			if (!Objects.equals(getBoundColumnText(), tStringMapping)) {
```
bpangburn, line 916: Per your suggestion.

errael (approved):
I looked at the Objects.equals code late summer, when I "got" that it was there. Nice code.

PR #64, opened 2021-01-12
cleanup warnings

errael:
Most of the issues are because AbstractComboBoxListSwingModel is not actually extend a swing model anymore since it uses a proxy. So no "serial" required.


bpangburn (approved):
Looks good.

PR #66, opened 2021-01-18
Add selectionPending property, to insure blank combo if new row.

errael:
Along with the stuff in SSBaseComboBox, there's a little tweak in SSDBNavImpl.

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -632,13 +632,13 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 * Control whether or not waiting for pending user selection. When selectionPending
 	 * is true, there is a nullItem and it is selected in the combo.
 	 * selectionPending is set to false automatically when a non nullItem is selected.
-	 * @param selectionPending true selects a possibly tempory nullItem
+	 * @param _selectionPending true selects a possibly tempory nullItem
 	 */
-	public void setSelectionPending(boolean selectionPending) {
-		if (this.selectionPending == selectionPending) {
+	public void setSelectionPending(boolean _selectionPending) {
+		if (selectionPending == _selectionPending) {
 			return;
 		}
-		this.selectionPending = selectionPending;
+		selectionPending = _selectionPending;
 
 		adjustForNullItem();
 		if (selectionPending) {
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -1003,7 +1003,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 			logger.debug(() -> String.format("%s: Setting to null.", getColumnForLog()));
 			setBoundColumnText(null);
 		} else {
-			logger.debug(String.format("%s: Setting to %s.",  getColumnForLog(), mapping));
+			logger.debug(() -> String.format("%s: Setting to %s.",  getColumnForLog(), mapping));
 			// TODO: need to avoid setting to same value
 			// for NavGroupState. Wonder why, avoids event?
 			//setBoundColumnText(String.valueOf(mapping));
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -37,11 +37,11 @@
  * ****************************************************************************/
 package com.nqadmin.swingset;
 
-// SSBaseComboBox.java
-
 import java.awt.Dimension;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.event.ItemEvent;
+import java.awt.event.ItemListener;
 import java.awt.event.KeyAdapter;
 import java.awt.event.KeyEvent;
 import java.io.Serializable;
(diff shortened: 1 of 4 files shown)


swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -336,6 +376,50 @@ public boolean hasSelection() {
...
+		return selectionPending;
+	}
+
+	/**
+	 * Control whether or not waiting for pending user selection. When selectionPending
+	 * is true, there is a nullItem and it is selected in the combo.
+	 * selectionPending is set to false automatically when a non nullItem is selected.
+	 * @param selectionPending true selects a possibly tempory nullItem
+	 */
+	public void setSelectionPending(boolean selectionPending) {
```
bpangburn, line 394: Stylistically, I'm not a fan of 'this.'

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -136,6 +136,24 @@
...
+
+			if (mapping == null) {
+				setBoundColumnText(null);
+				logger.debug(() -> String.format("%s: Setting to null.", getColumnForLog()));
+			} else {
+				setBoundColumnText(String.valueOf(mapping));
+				logger.debug(() -> String.format("%s: Setting to %s.",  getColumnForLog(), mapping));
+      			}
+*/			
+// BP gl-strict-contains code
```
bpangburn, line 156: @errael Any concerns about returning if we're dealing with a combo navigator (below) and keeping the above logic in updateRowset() around line 900? If not, we can leave the commented block for now in case there are unforeseen side effects and clean up later.

updateRowset() also has some of your event-bus prep code.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -264,16 +301,12 @@ protected BaseModel() {
 			BaseGlazedModel<M,O,O2> model = new BaseGlazedModel<>();
 			model.autoComplete = AutoCompleteSupport.install(_jc, model.getEventList(), null, model.getListItemFormat());
 			
-			model.autoComplete.setStrict(true);
 			model.autoComplete.setFilterMode(TextMatcherEditor.CONTAINS);
-			
+			model.autoComplete.setStrict(true);
```
bpangburn, line 305: You placed model.autoComplete.setStrict(true); after setFilterMode() so going with your version.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -380,6 +424,15 @@ public void setModel(ComboBoxModel<SSListItem> _model) {
 				: null;
 	}
 
+	/**
+	 * create SSBaseComboBox
+	 */
+	public SSBaseComboBox() {
```
bpangburn, line 430: Was thinking this was adding an empty constructor, but then remembered this is an abstract class.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -566,6 +619,50 @@ public boolean hasSelection() {
...
+			// Setting to true from false, select the nullItem;
+			// Events are a problem. After SSDBNavImpl.setSelectionPending
+			// the following ends up in SSBaseComboBoxListener.actionPerformed()
+			// then into setBoundColumnText then exception in RowSetOps.updateColumnText.
+			// TODO: merge this into adjustForNullItem, to avoid extra unregister/register
+			//
+			// Or better, with cleaner event model, like Navigator events,
+			// might be easier to avoid RowSetOps.updateColumnText.
+			// Actually, if there was an error badge next to the ComboBox
+			// (instead of those f*ing dialogs) then when AllowNull is false
```
bpangburn, line 654: We'll get there. I don't like the dialogs either.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1632,6 +1632,10 @@ public void updateSSComponent() {
 			if (!hasItems()) {
 				return;
 			}
+
```
bpangburn, line 1635: This method had already been scrubbed/shortened, but I guess you didn't have that code. setSelectionPending(true) seemed like the only new thing and I placed that before the call to getBoundColumnText();

swingset/src/main/java/com/nqadmin/swingset/SSDBNavImpl.java
```diff
@@ -147,6 +147,9 @@ protected void setComponents(final Container _container) {
             } else if(comps[i] instanceof JTextArea) {
             // IF IT IS A JTextArea, SET TO EMPTY STRING
                 ((JTextArea)comps[i]).setText("");
+            } else if (comps[i] instanceof SSBaseComboBox) {
```
bpangburn, line 150: Looks like setSelectionPending() takes care of setting to nullItem so setSelectionPending(true) is sufficient here.

bpangburn (commented):
@errael, please have a look at my commented/edits and see if you're in agreement or think further changes are needed.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: `setSelectionPending(true)` instead of `getEditor().setItem("")`

Oh, you had already gotten rid of `setItem("")`. Yes, I put it before `getBoundColumnText();`

swingset/src/main/java/com/nqadmin/swingset/SSDBNavImpl.java
errael: Right, at least that's the ide.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: So all of this (block above highlighted in green (and commented out by me in GitHub editor) was re-introduced into actionPerformed(). Based on your email, there should just be one line edited in updateRowset(). Will fix.

errael (commented):
AFAICT, you picked up the stuff, plus some contributions from the merge demons.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: "This method" is updateSSComponent() and you have to click the blue up/down arrow to expand it in review mode.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -905,7 +1003,7 @@ private void updateRowset() {
 			logger.debug(() -> String.format("%s: Setting to null.", getColumnForLog()));
 			setBoundColumnText(null);
 		} else {
-			logger.debug(String.format("%s: Setting to %s.",  getColumnForLog(), mapping));
+			logger.debug(() -> String.format("%s: Setting to %s.",  getColumnForLog(), mapping));
```
bpangburn, line 1006: This is where the logger.debug "() -> " fix needed to be after logic was moved out of the action listener. Again, I think that's the worst of it.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: Didn't think so because it seemed fine in my testing, but since you've actually been inside GL was going with your version.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: That's what I meant by empty constructor, but that might not be the right terminology. Doesn't matter as it should be a non-issue.

PR #67, opened 2021-01-21
Gl strict contains

bpangburn:
Using glazedlists 1.11.1201 from com.raelity.3rdparty.com.glazedlists with support for STRICT+CONTAINS.


errael (approved):
A quick scan. Looks good. (especially enjoy thinking about the glazed glitch commenting code disappearing)

PR #80, opened 2021-02-16
4.0.2 snapshot

### LINKED ISSUE #79: Consolidate Add/Remove of Component Listeners used for rowset binding
There is code in SSBaseComboBox.setSelectedMapping() that logs a warning and calls setSelectedItem(nullItem) if there is no matching item. There is a fairly common use-case where this is expected and that is where a SSDBComboBox is re-queried for every record due to a parent-child relationship. For example if we had one SSDBComboBox with companies and another one with employees, we would expected the employee combo to requery for each record where there was a change in company selected.

As soon as a navigation occurs, the component values are cleared and then the new value is loaded, but the combo has not been re-queried yet so there are no matches for setSelectedMapping() and we call setSelectedItem(nullItem). This is OK so long as the component listener used for binding is removed/disabled because the rowset will not get the null value. Later when the combo is requeried, the component will try to load the current column value from the rowset and this time setSelectedMapping() should succeed.

The "problem" is that the logger warning is scary because it indicates possible data corruption so we don't want to make the entry if the listener is off. I turned on debug level logging and expected to see the addition/removal of the listeners and I did not.

When my debugging was not as expected, I realized I was a little sloppy in how I was adding/removing the component value change listeners and was not funneling through common methods. The RowSet listeners are good, but the component value change listener implementations and listener type vary by component and should be consolidated.

One of the SS4 design goals was to funnel all RowSet listeners triggering component value changes through common methods. Same with all component value changes triggering RowSet updates. This also allows for checking that we don't accidentally create multiple change listeners for either the component or the RowSet. There can be other listeners, but for a given component binding, there should be exactly one rowset and one component listener that triggers a corresponding change in the value of the component or rowset. No duplicates. No cascading.

bpangburn:
Resolves #79 

Changes:
	1. SSDBComboBox - added constructor without _query parameter.
	2. SSFormViewScreenHelper API change, replacing:
	     protected abstract void retrieveAndSetNewPrimaryKey();
	   with:
	     protected abstract String retrieveNewPrimaryKey();
	3. Moved addSSComponentListener() and removeSSComponentListener() to
	   SSCommon, making companion methods in SSComponentInterface static.

Note that as I worked on the listener methods for several components, I found soft/hard tab issues so there will be some formatting changes. Main thing to check for each component is removal of add/removeSSComponentListener() methods and addition of getSSComponentListener().

### DIFF
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
@@ -245,10 +245,12 @@ public class Example4UsingHelper extends SSFormViewScreenHelper {
 	}
 
 	/**
-	 * Obtain and set the PK value for the new record.
+	 * Obtain and return the primary key value for the new record.
 	 */
 	@Override
-	protected void retrieveAndSetNewPrimaryKey() {
+	protected String retrieveNewPrimaryKey() {
+		
+		String newPrimaryKey = null;
 
 		try {
 
@@ -257,8 +259,7 @@ public class Example4UsingHelper extends SSFormViewScreenHelper {
 					.executeQuery("SELECT nextval('part_data_seq') as nextVal;");
 			rs.next();
 			final int partID = rs.getInt("nextVal");
-			//txtPartID.setText(String.valueOf(partID));
-			getTxtPrimaryKey().setText(String.valueOf(partID));
+			newPrimaryKey = String.valueOf(partID);
 			rs.close();
 			
 		} catch(final SQLException se) {
@@ -266,6 +267,8 @@ public class Example4UsingHelper extends SSFormViewScreenHelper {
 		} catch(final Exception e) {
 			logger.error("Exception occured initializing new record.",e);
 		}
+		
+		return newPrimaryKey;
 
 	}
 
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
@@ -238,10 +238,7 @@ public class Example4UsingHelper extends SSFormViewScreenHelper {
 	@Override
 	protected void populateSSComboBoxes() {
 		// SET COMBO OPTIONS
-		//cmbPartColor.setOptions(new String[] { "Red", "Green", "Blue" });
-		System.out.println(cmbPartColor);
-		cmbPartColor.setOptions(Arrays.asList(new String[] { "Red", "Green", "Blue" }));
-		
+		cmbPartColor.setOptions(new String[] { "Red", "Green", "Blue" });
 	}
 
 	/**
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -708,7 +708,17 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 				// nullItem is either special first list item
 				// or it is null. It is null when getAllowNull() is false
 				item = nullItem;
-				logger.warn(String.format("%s: No mapping available for %s in combobox, setSelectedItem(null)", getColumnForLog(), _mapping));
+				// BP_2021-02-16:
+				// We expect to get here if we have a child combo where the contents are requeried on each record.
+				// As soon as a navigation occurs, the component values are cleared and then the new value is loaded,
+				// but the combo has not been re-queried yet so there are no matches for setSelectedMapping() and we
+				// call setSelectedItem(nullItem). This is OK so long as the component listener used for binding
+				// is removed/disabled because the rowset will not get the null value. Later when the combo is
+				// requeried, the component will try to load the current column value from the rowset and this time
+				// setSelectedMapping() should succeed.
+				if (getSSCommon().isSSComponentListenerAdded()) {
+					logger.warn(String.format("%s: No mapping available for %s in combobox, setSelectedItem(null)", getColumnForLog(), _mapping));
+				}
 			}
 			setSelectedItem(item);
 			
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -331,11 +331,6 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 */
 	protected SSCommon ssCommon = new SSCommon(this);
 
-	/**
-	 * Component listener.
-	 */
-	protected SSBaseComboBoxListener ssBaseComboBoxListener;
-
 //	/**
 //	 * List item used to workaround GlazedList CONTAINS bug/glitch.
 //	 */
@@ -856,10 +851,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 */
 	@Override
 	public SSBaseComboBoxListener getSSComponentListener() {
-		if (ssBaseComboBoxListener==null) {
-			ssBaseComboBoxListener = new SSBaseComboBoxListener();
-		}
-		return ssBaseComboBoxListener;
+		return new SSBaseComboBoxListener();
 	}
 
 	/**
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -37,6 +37,9 @@
  * ****************************************************************************/
 package com.nqadmin.swingset;
 
+import static com.nqadmin.swingset.models.AbstractComboBoxListSwingModel.addEventLogging;
+import static com.nqadmin.swingset.models.OptionMappingSwingModel.asOptionMappingSwingModel;
+
 import java.awt.Dimension;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
@@ -66,9 +69,6 @@ import com.nqadmin.swingset.utils.SSComponentInterface;
 import ca.odell.glazedlists.BasicEventList;
 import ca.odell.glazedlists.matchers.TextMatcherEditor;
 import ca.odell.glazedlists.swing.AutoCompleteSupport;
-
-import static com.nqadmin.swingset.models.AbstractComboBoxListSwingModel.addEventLogging;
-import static com.nqadmin.swingset.models.OptionMappingSwingModel.asOptionMappingSwingModel;
 
 // SSBaseComboBox.java
 //
@@ -334,7 +334,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	/**
 	 * Component listener.
 	 */
-	protected final SSBaseComboBoxListener ssBaseComboBoxListener = new SSBaseComboBoxListener();
+	protected SSBaseComboBoxListener ssBaseComboBoxListener;
 
 //	/**
 //	 * List item used to workaround GlazedList CONTAINS bug/glitch.
@@ -840,6 +840,17 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	public SSCommon getSSCommon() {
 		return ssCommon;
 	}
+	
+	/**
+	 * {@inheritDoc }
+	 */
+	@Override
+	public SSBaseComboBoxListener getSSComponentListener() {
+		if (ssBaseComboBoxListener==null) {
+			ssBaseComboBoxListener = new SSBaseComboBoxListener();
+		}
+		return ssBaseComboBoxListener;
+	}
 
 	/**
 	 * {@inheritDoc }
@@ -848,25 +859,6 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	public void setSSCommon(final SSCommon _ssCommon) {
 		ssCommon = _ssCommon;
 
-	}
-
-	/**
-	 * Adds any necessary listeners for the current SwingSet component. These will
-	 * trigger changes in the underlying rowset column.
-	 */
-	@Override
-	public void addSSComponentListener() {
-		addActionListener(ssBaseComboBoxListener);
-
-	}
-
-	/**
-	 * Removes any necessary listeners for the current SwingSet component. These
-	 * will trigger changes in the underlying RowSet column.
-	 */
-	@Override
-	public void removeSSComponentListener() {
-		removeActionListener(ssBaseComboBoxListener);
 	}
 
 	/**
@@ -1002,7 +994,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 */
 	private void updateRowset() {
 		
-		removeRowSetListener();
+		ssCommon.removeRowSetListener();
 		
 		M mapping = getSelectedMapping();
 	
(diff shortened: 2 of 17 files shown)


swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
```diff
@@ -245,10 +245,12 @@ protected void populateSSComboBoxes() {
 	}
 
 	/**
-	 * Obtain and set the PK value for the new record.
+	 * Obtain and return the primary key value for the new record.
 	 */
 	@Override
-	protected void retrieveAndSetNewPrimaryKey() {
+	protected String retrieveNewPrimaryKey() {
```
bpangburn, line 251: This seemed like a cleaner way to handle in practice.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -37,6 +37,9 @@
  * ****************************************************************************/
 package com.nqadmin.swingset;
 
+import static com.nqadmin.swingset.models.AbstractComboBoxListSwingModel.addEventLogging;
```
bpangburn, line 40: I used the "organize imports" command in Eclipse if I was left with unnecessary imports following coding changes. It changes the sorting.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -708,7 +708,17 @@ public void setSelectedMapping(final M _mapping) {
 				// nullItem is either special first list item
 				// or it is null. It is null when getAllowNull() is false
 				item = nullItem;
-				logger.warn(String.format("%s: No mapping available for %s in combobox, setSelectedItem(null)", getColumnForLog(), _mapping));
+				// BP_2021-02-16:
```
bpangburn, line 711: This is what triggered the listener rewrite.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -1002,7 +1004,7 @@ protected void adjustForNullItem() {
 	 */
 	private void updateRowset() {
 		
-		removeRowSetListener();
+		ssCommon.removeRowSetListener();
```
bpangburn, line 1007: There is also a static method in SSComponentInterface so I could have called SSComponentInterface.removeRowSetListener(this), but the call to SSCommon seemed cleaner in practice.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -1036,12 +1038,16 @@ private void updateRowset() {
...
+	 * <p>
 	 * This method is the logical inverse of {@link #registerAllActionListeners}.
+	 * <p>
+	 * Note that the ActionListener used for binding will be removed without
+	 * passing through SSCommon.removeSSComponentListener()
 	 * 
 	 * @param comboBox combo box from which to remove listeners
 	 * @return array of ActionListeners removed from combo box (for adding back later)
 	 */
+	// TODO: Consider passing in SSBaseComboBox and identifying SSBaseComboBoxListener
```
bpangburn, line 1050: Since register/unregisterAllActionListeners() is a temporary thing, I don't think there is any impact on the binding so decided to leave them alone.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -321,6 +321,24 @@ public SSDBComboBox() {
 		listItemFormat.setPattern(JDBCType.DATE, dateFormat);
 		listItemFormat.setSeparator(separator);
 	}
+
```
bpangburn, line 324: New constructor for SSDBComboBox that does not require the query. Query isn't executed until later and my screen  helper isolates setting and execution of queries into a single method for a screen so it was redundant to pass the query into the Constructor.

swingset/src/main/java/com/nqadmin/swingset/SSImage.java
```diff
@@ -72,60 +72,60 @@
  */
 public class SSImage extends JPanel implements SSComponentInterface {
```
bpangburn, line 74: So here is one with lots of soft to hard tab conversions. Sorry.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -395,31 +419,64 @@ public SSCommon(final SSComponentInterface _ssComponent) {
...
+			rowSetListenerAdded = true;
+			logger.debug(boundColumnName + " - RowSet Listener added.");
 		}
 	}
 	
 	/**
 	 * Method to add any SwingSet Component listener(s).
 	 */
-	public void addSSComponentListener() {
+	public final void addSSComponentListener() {
```
bpangburn, line 446: **So if you study one thing carefully in this PR, this method should be that thing. And the removeSSComponentListener() counterpart.**

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -670,33 +759,65 @@ protected void init() {
...
-
+	
 	/**
-	 * Class to remove a Document listener when the SwingSet component is a
-	 * JTextComponent
+	 * Method to add any SwingSet Component listener(s).
 	 */
-	public void removeDocumentListener() {
-		((javax.swing.text.JTextComponent) getSSComponent()).getDocument().removeDocumentListener(documentListener);
+	public final void removeSSComponentListener() {
```
bpangburn, line 766: **Check me out too please.**

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -395,31 +419,64 @@ public SSCommon(final SSComponentInterface _ssComponent) {
...
+			} else if (ssComponent instanceof JTextComponent) {
+				addSSDocumentListener();
+			} else {
+				// DIPLAY WARNING FOR UNKNOWN EVENT LISTENER
+				String message = String.format("%s - Encountered unknown Component Event Listener for: %s. Unable to add component listener.",
+						getColumnForLog(), ssComponent.getClass().getSimpleName());
+				logger.error(message);
+				JOptionPane.showMessageDialog((JComponent)getSSComponent(), message, "Unknown Component Event Listener", JOptionPane.ERROR_MESSAGE);
+			}
 		}
```
bpangburn, line 476: **So if you study one thing carefully in this PR, this method should be that thing. And the removeSSComponentListener() counterpart.**

errael: > this method should be that thing. And the removeSSComponentListener() counterpart.

I haven't looked yet, but is that "one thing" `addSSDocumentListener`? And is `removeSSComponentListener` the right name?

bpangburn: Looking for a sanity check on SSCommon.addSSComponentListener() and SSCommon.removeSSComponentListener()

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
errael: > TODO: Consider passing in SSBaseComboBox and identifying SSBaseComboBoxListener

I don't understand what this comment, for both methods, is about. What's the goal? (and it's been a long time)

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
errael: Any reason to deprecate the constructor with the query?

swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
errael: This is beyond me.

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
bpangburn: I was contemplating isolating/identifying and removing the component listener used for rowset binding separately and pushing those calls through SSCommon for logging/tracking. Since this is just done momentarily and  then added back, I don't think it's worth the effort involved.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: Thought about it, but nah. It doesn't hurt anything and if you're not using the form helper classes then it doesn't really matter where you set the query.

swingset/src/main/java/com/nqadmin/swingset/utils/SSComponentInterface.java
```diff
@@ -330,36 +366,63 @@ default RowSet getRowSet() {
...
+	 * Indicates if the components RowSet listener is added/enabled.
+	 * 
+	 * @param _component SwingSet component to be checked for a RowSet listener
+	 *
+	 * @return true if the component's RowSet listener is added/enabled, otherwise false
 	 */
-	void removeSSComponentListener();
+	static boolean isRowSetListenerAdded(SSComponentInterface _component) {
+		return _component.getSSCommon().isRowSetListenerAdded();
+	}
```
errael, line 388: Why not
```
boolean isRowSetListenerAdded() {
    return getSSCommon().isRowSetLIstenerAdded();
```

And I guess there's a few others. Probably something I'm missing.

swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
bpangburn: When you add a new record to the database, the user can populate the primary key column if there is one or you can rely on a unique index provided by the database. Previously in my form helper class I had a abstract method to say "take care of the primary key for the new record". In the form helper, I actually maintain an SSTextField bound to the primary key column so I thought it would be better to change the API to just request the new value and let the helper take care of setting it. Eliminates a little code and hopefully makes the helper more helpful.

swingset/src/main/java/com/nqadmin/swingset/utils/SSFormViewScreenHelper.java
```diff
@@ -873,12 +881,20 @@ protected void updateScreen() {
...
+	 * Generally there will be calls to cmbMySSDBCombo.setQuery(myNewCmbSQL); and
+	 * cmbMySSDBCombo.execute();
+	 * <p>
+	 * If the SSDBComboBox query is unchanged regardless of the rowset, then it is
+	 * sufficient to just call cmbMySSDBCombo.execute();
+	 * <p>
+	 * It is REDUNDANT/UNNECESSARY to call cmbMySSDBCombo.execute() here and in
+	 * bindComponents().
 	 */
 	protected abstract void updateSSDBComboBoxes();
```
errael, line 900: First time I've looked into a screen help file. Looks like this situation was already covered, at least partly :-)

swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example4UsingHelper.java
bpangburn: The idea is to query the database for the next value in the appropriate sequence (part_data_seq for our part_data table) and return the result.
`executeQuery("SELECT nextval('part_data_seq') as nextVal;");`

errael (commented):
A few places where clarifications might be good, but I don't see anything that would necessarily be a problem.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
bpangburn: Good point. I added JavaDoc. Not sure how I would enforce the user to stash a single listener in SSCommon?

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -395,31 +419,64 @@ public SSCommon(final SSComponentInterface _ssComponent) {
...
 	/**
 	 * Method to add any SwingSet Component listener(s).
 	 */
-	public void addSSComponentListener() {
+	public final void addSSComponentListener() {
 		if (!ssComponentListenerAdded) {
-			ssComponent.addSSComponentListener();
-			ssComponentListenerAdded = true;
+			
+			Object eventListener = getSSComponent().getSSComponentListener();
```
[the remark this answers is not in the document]
bpangburn: Fixed.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -395,31 +419,64 @@ public SSCommon(final SSComponentInterface _ssComponent) {
...
+			} else {
+				// DIPLAY WARNING FOR UNKNOWN EVENT LISTENER
+				String message = String.format("%s - Encountered unknown Component Event Listener for: %s. Unable to add component listener.",
+						getColumnForLog(), ssComponent.getClass().getSimpleName());
+				logger.error(message);
+				JOptionPane.showMessageDialog((JComponent)getSSComponent(), message, "Unknown Component Event Listener", JOptionPane.ERROR_MESSAGE);
+			}
 		}
+
+		ssComponentListenerAdded = true;
```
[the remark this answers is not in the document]
bpangburn: Cleaned this up in both methods presuming success, but setting the opposite if the default case is encountered.

swingset/src/main/java/com/nqadmin/swingset/utils/SSComponentInterface.java
bpangburn: Struggled with this. It's an interface so I can't make a method final, only default so that can invite the user to override. Making it static, they can't. I want to be as restrictive as possible. Could probably even eliminate the static methods.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -670,33 +759,65 @@ protected void init() {
...
-	 * Method to remove any SwingSet Component listener(s).
+	 * Method to remove the RowSet listener.
 	 */
-	public void removeSSComponentListener() {
-		if (ssComponentListenerAdded) {
-			ssComponent.removeSSComponentListener();
-			ssComponentListenerAdded = false;
+	public final void removeRowSetListener() {
+		// rowSetListenerAdded==true indicates that rowset is not null, so not checking
+		// rowSetListenerAdded==true indicates that rowSetListener is not null, so not checking
```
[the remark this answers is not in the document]
bpangburn: Good point. Added the null rowset check. Wondering what else a random setRowSet(null) would break!

swingset/src/main/java/com/nqadmin/swingset/utils/SSComponentInterface.java
errael: Got it, makes sense. It makes the intention explicit.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
errael: > enforce the user to stash a single listener in SSCommon?

Do it yourself?
```
private final EventListener eventListener;
public SSCommon(final SSComponentInterface _ssComponent) {
    eventListener = _ssComponent.getSSComponentListener();
```

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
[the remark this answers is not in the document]
errael: I know. Could always do
```
public void setRowSet(final RowSet _rowSet) {
    Ojbects.requiredNonNull(_rowSet);
```

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
bpangburn: OK. So you'd still have getSSComponentListener(), but it would become single use and the implementation could probably be simplified to return new SSXYZComponentListener();

Issue #85, opened 2021-03-01
SSDBComboBox.updateOption() not always updating combo editor text


bpangburn:
Demo->Example 4 With Helper:

Added a focus lost listener to Part Name to update the combo navigator via a call to SSDBComboBox.updateOption() if the Part Name changes. Working as expected for the first record where part name is edited, but not on other rows. Sometimes a navigation takes place bringing the user back to the first row.

Steps to replicate:
1. Open Example 4 Using Helper
2. Change Part Name from "Nut" to "Nutty" and tab to next field. Navigator updates.
3. Move to 2nd record.
4. Change Part Name from "Bolt" to "Bolty" and tab to next field. Navigator fails to update.
5. Navigate to record 3 then back to 2. Update is there.

Doing a refresh on the data navigator resets things.

bpangburn:
@errael was able to fix SSDBComboBox.updateOption() with the following modifications.

```
	public boolean updateOption(final Long _mapping, final String _option) {

		boolean result = false;

		try (Model.Remodel remodel = optionModel.getRemodel()) {
			final int index = remodel.getMappings().indexOf(_mapping);
			if (index >= 0) {
				boolean isSelectedItem = Objects.equals(_mapping, getSelectedMapping());
				remodel.setOption(index, _option);
				result = true;
				if (isSelectedItem) {
					// Modifying the underlying list item that corresponds to the
					// current selection; strict glazed may change the selection.
					// Select the modified item so the same mapping is selected.
					SSListItem item = remodel.get(index);
					setSelectedItem(item);
				}
// TODO Confirm that eventList is not reordered by GlazedLists code.
			}
		} catch (final Exception e) {
			logger.error(getColumnForLog() + ": Exception.", e);
		}
		
		if (!result) {
			logger.warn(() -> String.format("%s: Unable to update Mapping of [%s] with Option of '%s'.",
				getColumnForLog(), _mapping, _option));
		}

		return result;
	}
```
@errael also opened an issue in GlazedLists with further insights:
https://github.com/glazedlists/glazedlists/issues/702

bpangburn:
Fixed in swingset-4.0.4

Issue #46, opened 2020-12-30
Move updateSSComponent() from combo implementations in SSBaseComboBox


bpangburn:
Currently there are similar, but separate implementations of updateSSComponent() in SSDBComboBox and SSComboBox. They should be consolidated in SSBaseComboBox, but have to handle the different types expected or make it type-agnostic.

bpangburn:
Somewhat related to Issue #17

bpangburn:
Per @errael:


Is this an aspect of #17? I'd consider splitting RowSetOps.updateColumnText, something like

```
Object RowSetOps.parseText(JDBCTYpe type, String text)

RowSetOps.updateColumnText(...)
    _rowSet.updateObject(parseColumnText(...))
```

And then (M)RowSetOps.parseText(...) could work (possible with a check warning).

Adding an abstractMethod like (M)parseString is certainly a workaround.

errael:
@bpangburn What do you think about a `Object RowSetOps.parseText(JDBCTYpe type, String text)` method?

I guess there would be a method `RowSetOps.parseText(int columnIndex, String text)` and it picks up the JDBCType from the metadata. This method would invoke the `parseText(type, text)` method.

This counts on the database type matching up with what's needed. BIGINT for Long in this case to get a `Mapping`. 

If that does't work, either need to explicitly tell the combo box, or use some refection to get the proper type. (guava has some stuff to help pick up generic type info, https://stackoverflow.com/questions/3403909/get-generic-type-of-class-at-runtime).

bpangburn:
@errael I think this is a good approach. While it does not yet use JDBCTypes, SSFormattedTextField.updateSSComponent() does something somewhat similar except that it expects the component can handle the Object type and uses getRowSet().getObject(columnName) so it doesn't cast back/forth to a String.

errael:
Tried a few things, note in the following `getMappingType()` returns the class of *(M)*
1. `M t = (M) parseText(getBoundColumnJDBCType(), boundColumnText);`
This uses the metadata as a guide for parsing the text into an object. This fails since SSDBCombo always expects a Long now matter the database type.
2. `M t = (M)RowSetOps.getValueOf(getMappingType(), boundColumnText);`
This one uses the class of the mapping and if there is a `valueOf(String)` method, invokes it. This is similar to how the combo editor reverse maps a string to an item. Limited and unsatisfying.
3. `M t = (M)parseText(classToJdbcType(getMappingType()), boundColumnText);`
[...]

PR #86, opened 2021-03-18
4.0.4 snapshot

bpangburn:
1. SSBaseComboBox.updateOption(). Combo editor was not updating consistently. See https://github.com/bpangburn/swingset/issues/85
2. Updated Log4j and JUnit Maven dependencies.
3. Added explicit Maven plugins versions to eliminate warnings.
4. Initial fix for Issue #46 moving updateSSComponent out of SSComboBox and SSDBComboBox.


swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -1025,6 +1025,73 @@ private void updateRowset() {
...
+					targetValue = (M)(Object)Integer.parseInt(boundColumnText);
+				} else if (this instanceof SSDBComboBox) {
+					targetValue = (M)(Object)Long.parseLong(boundColumnText);
+				} else {
+					throw new Exception();
+				}
+			}
+			
+			// LOGGING
+			logger.debug("{}: targetValue - " + targetValue, () -> getColumnForLog());
```
errael, line 1080: Patch on the way to minimize unchecked scope and avoid creating StringBuilder when not logging

errael (commented):
Only really looked at base combo.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1484,16 +1481,29 @@ public void setSeperator(final String _separator) {
...
 		boolean result = false;
 
 		try (Model.Remodel remodel = optionModel.getRemodel()) {
 			final int index = remodel.getMappings().indexOf(_mapping);
 			if (index >= 0) {
+				boolean isSelectedItem = Objects.equals(_mapping, getSelectedMapping());
 				remodel.setOption(index, _option);
 				result = true;
+				// TODO: Possible that this block and isSelectedItem variable
+				// may be eliminated if a bug is discovered/fixed in GlazedLists.
```
errael, line 1496: Changing what's in the ComboEditor, which may be done indirectly when modifying the current item, might change the currently selected item when STRICT. So something is needed to insure that the selected mapping before the change is the selected mapping after the change. Otherwise the first item in the list becomes selected (as has been seen, and I guess that's what kicks the navigator?)

Notice that the bug https://github.com/glazedlists/glazedlists/issues/702 affects subsequent programmatic changes to the underlying EventList, not the first one.

errael (commented):
A couple of comments on the comments. Which may, or may not, be entirely correct, but probably something to consider.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1484,16 +1481,29 @@ public void setSeperator(final String _separator) {
 	 */
```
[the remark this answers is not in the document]
bpangburn: Agree that this comment seems unnecessary. Best practice would be to use SSSyncManager. If a developer is coding up a custom combo navigator then they can make whatever decisions/assumptions they want about when to save changes to a row. I'll push an update removing the comment.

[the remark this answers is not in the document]
bpangburn: Also added the SSSyncManager tweak to the CHANGELOG.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
bpangburn: @errael are you saying that we would want to retain the `if (isSelectedItem) {` check regardless of whether https://github.com/glazedlists/glazedlists/issues/702 turns out to be a bug and is fixed? In that case we should just eliminate the TODO and possibly replace with a comment similar to what you wrote above?

errael: > are you saying that we would want to retain the `if (isSelectedItem) {` check

Yes, that's what I meant. My assumption is that the mapping that was selected before `updateOption` is called should be the mapping after the call. Typically, if the selected item is changed, glazed will change the selection.

>  eliminate the TODO

That works. The comment that's there may need adjustment/clarification.

If, `updateOption` doesn't need to maintain the selected mapping, then the `if` could be removed after the glazed bug is fixed. I don't recall exactly, but I think the `setSelectedItem(item)` is what avoid the glazed bug on the *next* change.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1469,12 +1466,6 @@ public void setSeperator(final String _separator) {
 	 * <p>
 	 * If more than one item is present in the combo for that mapping, only the
 	 * first one is changed.
-	 * <p>
```
bpangburn, line 1472: @errael See JavaDoc removed per code review.

swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -1484,16 +1475,37 @@ public void setSeperator(final String _separator) {
...
+		// updating of combo editor. See https://github.com/bpangburn/swingset/issues/85
+		
 		boolean result = false;
 
 		try (Model.Remodel remodel = optionModel.getRemodel()) {
 			final int index = remodel.getMappings().indexOf(_mapping);
 			if (index >= 0) {
+				boolean isSelectedItem = Objects.equals(_mapping, getSelectedMapping());
 				remodel.setOption(index, _option);
 				result = true;
```
bpangburn, line 1488: @errael See updated comments and removal of TODO per code review.

errael: Cool

PR #94, opened 2021-10-22
4.0.5 snapshot

bpangburn:
I want to do a point release that picks up the latest JdbcRowSetImpl along with some other POM plugin and dependency updates. Then I'll try to pick back up on the formatted text field classes.

Note you'll also see some formatted text field code that was added and then removed because I accidentally committed to the wrong branch.


swingset-demo/pom.xml
```diff
@@ -22,11 +22,11 @@
 		<version.java>1.8</version.java>
 
 		<version.log4j>2.14.1</version.log4j>
-		<version.swingset>4.0.4</version.swingset>
+		<version.swingset>4.0.5-SNAPSHOT</version.swingset>
```
errael, line 25: After doing (from parent directory) cd swingset; mvn clean; mvn install. I got an error in swingset-demo. I couldn't find 
`com.nqadmin.swingset:swingset:jar:4.0.5-SNAPSHOT`

Got rid of this "-SNAPSHOT" and things went OK.

errael (changes requested):
With the change so that demo doesn't use the snapshot, built and ran demo with-dependencies. Brought up a few screens; didn't try anything else.

I notice that the enforcer only allows specific versions, and not version X and later.

swingset-demo/pom.xml
bpangburn: Removed -SNAPSHOT as suggested.

errael (approved):
I'm mostly assuming all the newer versions are cool. Wiped 4.0.5* from local repository, updated local clone, built, ran.

Issue #95, opened 2021-11-18
Exception when inserting new record on empty rowset


bpangburn:
If a user has a screen based on an empty rowset, when they try to add a record, the SSDataNavigator attempts to autosave with a call to private method commitChanges(). @prasanthreddy-git  proposed the following fix:

```
			if (rowSet.getRow() > 0) {
				doUpdateRow();
			} 
```

errael:
I wonder which exception was thrown, and from where. Javadoc for updateRow doesn't explicitly call out this case as an exception; but then again can't update a row that doesn't exist so probably a javadoc deficiency.

Interesting to note that with a smart NavBar implementation, this wouldn't be called since nothing is dirty. At least that's what would be expected.

PR #96, opened 2021-11-18
4.0.5 snapshot

bpangburn:
Three changes:
1. bug fix for Issue #95 
2. updated the dependency-check-maven version in the POMs
3. updated changelog


swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
```diff
@@ -902,7 +902,12 @@ private boolean commitChanges() throws SQLException {
 			}
 			//rowSet.updateRow();
 			//dBNav.performPostUpdateOps();
-			doUpdateRow();
+
+			// 2021-11-18: prasanthreddy-git noticed that we didn't check for an empty
+			// rowset before trying to save changes
+			if (rowSet.getRow() > 0) {
+				doUpdateRow();
+			}
```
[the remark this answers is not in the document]
prasanthreddy-git: Looks like we are calling performPostUpdateOps in doUpdateRow which should not apply if the update is not needed. So either place should be fine. Would be better if we bring the call to rowset update & postUpdateOps itself into commitChanges as doUpdateRowWithoutPostUpdateOps & doUpdateRow don't have any logic in them.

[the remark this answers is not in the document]
errael: (the logic I was referring to was calling postUpdateOps)

There was a proposed change for smart NavBar, see #40, that didn't make it into 4.0.

Some of the changes from that proposal were brought in to 4.0; particularly those that localized setting the enable/disable state of the NavBar buttons rather than having that logic scattered around the actions since that tended to be buggy. While doing that the method `commitChanges()` was introduced which subsumed `doUpdateRow()`. So yeah, now I agree that the _fix belongs in, or around, `commitChanges()`_ and **`doUpdateRow()` can be removed**.

I'd be inclined to change the name of `commitChanges()` to `commitChangesIfAllowed()`; with the name change, the conditional return which always appears when calling `commitChanges()` is easier to understand. And I'd probably change `doUpdateRowWithoutPostUpdateOps()` to something like `commitChangesIfAllowedWithoutPostUpdateOps()`, maybe a little less verbose but to something so the update methods sort together. With these name changes, the recommended fix would go in `commitChangesIfAllowedWithoutPostUpdateOps()`.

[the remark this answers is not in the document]
bpangburn: Let me take a stab at this and commit an updated SSDataNavigator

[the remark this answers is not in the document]
bpangburn: @errael  @prasanthreddy-git  Have a look at the SSDataNavigator changes (pushed to this merge request) and see if they are to your liking. Tested a bit and seems to be working as expected. Hopefully code is a bit cleaner now.

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
```diff
@@ -886,23 +876,48 @@ protected void addToolTips() {
...
+	 * @throws SQLException SQL Exception if rowset call to updateRow() fails
+	 */
+	private boolean commitChangesIfAllowed() throws SQLException {
+		
+		boolean result = commitChangesIfAllowedWithoutPostUpdateOps();
+		
+		// After successful commit, perform any post-update operations
+		if (result) {
+			dBNav.performPostUpdateOps();
+		}
```
errael, line 894: Just to note: `dBNav.performPostUpdateOps()` is executed even if there are no rows in result set. I'd guess that doesn't matter.

errael (commented):
Change looks good to me.

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
bpangburn: Good point. I think this is how it worked previously, but if you look at the JavaDoc for the default performPostUpdateOps(), it says, "Method to perform operations after the updateRow has been called." Well, updateRow() has not been called if there aren't any rows so I'm inclined to change this.

bpangburn: @errael  Please have a look at my changes to commitChangesIfAllowedWithoutPostUpdateOps(). Now returns false if there are no records.

It still returns true if modification==false, but I guess that's OK. Feels like some cleanup is needed across SSDBNav.allowUpdate() and SSDataNavigator.setModification()/getModification().

errael: The `modification` variable and it's `get/set` methods stuff are inherited from pre 4.0. IIRC from my brief investigation working on 4.0, there was some situation where it was impermissible to modify the database; it may have even been for a specific database vendor. It may be that now this can be thrown out.

About returning false if there are no records. That means a bunch of code that used to be executed would no longer be executed, not just `performPostUpdateOps()`. Similarly, returning true if modification is false is pretty different. 

At this point in 4.0, I'd be inclined to make a minimal change to fix the bug. Then clean this area up in the next release.

bpangburn: @errael @prasanthreddy-git 

Probably went too far, but decided to merge the two methods and add an argument to control post update operations.
Previously, if modification==false, commitChanges() returned true so I don't think we're changing that behavior.

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
```diff
@@ -888,24 +874,48 @@ protected void addToolTips() {
...
+	 * modifications are allowed. After committing, it performs any
+	 * post-update operations.
+	 * <p>
+	 * If modification==false, then skip the update and return as
+	 * successful, unless we have an empty rowset. 
+	 * 
+	 * @param _performPostUpdateOps true if performPostUpdateOps() should
+	 * 	be called after successful update, otherwise false
+	 * 
 	 * @return true unless dBNav.allowUpdate() returns false
```
errael, line 886: I don't think this comment is accurate anymore.

errael (commented):
Looks good, clear and much simpler.

swingset/src/main/java/com/nqadmin/swingset/SSDataNavigator.java
```diff
@@ -704,7 +694,10 @@ private void addNavListeners() {
 			logger.debug("ADD button clicked.");
 			removeRowsetListener();
 			try {
-				if (!commitChanges()) return;
+				// Only attempt to commit to database if we are on a row (non-empty rowset)
+				if (rowSet.getRow() > 0) {
+					if (!commitChangesToDatabase(true)) return;
+				}
```
errael, line 700: These three lines can be replaced by
```
commitChangesToDatabase(true);
```
No conditional, it's basically a nop when `rowSet.getRow() == 0`.

I had trouble figuring out what was going on, why wasn't it ok. I guess this is an example of where treating empty data set the same as `dBNav.allowUpdate()` is an issue. After further thought, I'm wondering if
```
if (!dBNav.allowUpdate()) return;
```
should be in this method. Previously, if `dBNav.allowUpdate()` was false, the insert row could not be opened.

bpangburn: Agree. That row check became redundant with recent changes. Pushed the change.

dBNav.allowUpdate() is true by default. Guess there could be instances where a developer overrides it to false (which is presumably why the SSDBNav class is there), but that class is meant to be actually stop the user from doing things whereas the get/setModification() method and class variable allow the user to proceed as though things are actually persisted to the database (but they aren't). I think of modification==false as being sort of a demo mode.

errael: So I can't find anything else to nit pick about :-)

I finally looked at SSDBNav, I misunderstood some things. I still don't entirely understand it, but I have a better idea.

There's the perform ops, they're pretty clear; IIRC they're for workarounds for problems with databases. It seems they wouldn't change after app startup and connecting with the database.

Then there's the allow* methods. I'm confused how they might be used. I was wondering if they might be for dynamic changes depending on things that are going on. Or are they an escape hatch for yet unforseen problems?

prasanthreddy-git:
The allowXXXX methods are meant for more fine grained control (dynamic decision making based on each row).

errael:
> The allowXXXX methods are meant for more fine grained control (dynamic decision making based on each row).

If they're ever needed their signature might need to be revisited. I would think that at least knowing the rowSet would be important. Or maybe the idea is that side channels need to be used to maintain state.

PR #123, opened 2022-03-02
Work around log4j LogManager.getLogger() issue with gui builder

errael:
This is a quick 'n easy fix to the problem. The exact source of the problem is still under investigation.

This could be a problem in NB and not in log4j. See https://issues.apache.org/jira/browse/LOG4J2-3420.


bpangburn (approved):
Seems fine. Updated CHANGELOG. Added a comment to SSUtils.getLogger() to cite this PR.

Issue #133, opened 2022-04-29
SSList causes a NullPointerException when value in database is null


prasanthreddy-git:
SSList causes a NullPointerException when the value in the database is null

java.lang.NullPointerException: null
at com.nqadmin.swingset.models.SSDbArrayModel.readData(SSDbArrayModel.java:90) ~[swingset-4.1.0-SNAPSHOT.jar:?]
at com.nqadmin.swingset.SSList.updateSSComponent(SSList.java:539) ~[swingset-4.1.0-SNAPSHOT.jar:?]
at com.nqadmin.swingset.utils.SSCommon.updateSSComponent(SSCommon.java:1039) ~[swingset-4.1.0-SNAPSHOT.jar:?]
at com.nqadmin.swingset.utils.SSCommon.bind(SSCommon.java:526) ~[swingset-4.1.0-SNAPSHOT.jar:?]
at com.nqadmin.swingset.utils.SSCommon.bind(SSCommon.java:554) ~[swingset-4.1.0-SNAPSHOT.jar:?]
at com.nqadmin.swingset.utils.SSCommon.bind(SSCommon.java:568) ~[swingset-4.1.0-SNAPSHOT.jar:?]
at com.nqadmin.swingset.utils.SSComponentInterface.bind(SSComponentInterface.java:130) ~[swingset-4.1.0-SNAPSHOT.jar:?]

prasanthreddy-git:
Checking if the object list is null before converting it into an array

bpangburn:
Should be fixed in SwingSet 4.0.10

Issue #132, opened 2022-04-29
SSCheckBox throws a NullPointerException when column value is null


prasanthreddy-git:
SSCheckBox is throwing a NullPointerException when the column value in the database is null

prasanthreddy-git:
Modified the way the equals check is performed with predefined constants against the database value

bpangburn:
Should be fixed in SwingSet 4.0.10

Issue #131, opened 2022-04-29
SSComboBox causes IllegalArgumentException


prasanthreddy-git:
When the display field column is of type date and the value in the database is null you get the following exception.
java.lang.IllegalArgumentException: Cannot format given Object as a Date

prasanthreddy-git:
When the date/time value from the database is null not trying to format it. Not adding anything to the display either, if not a "null" string would be appended to the display text

bpangburn:
Should be fixed in SwingSet 4.0.10

PR #135, opened 2022-04-30
Updated code to fix null pointer exceptions in SSListBox and SSCheckBox.

bpangburn:
@errael these fixes look good to me. However, since 2 of the 3 fixes were in the models package, I thought it would be good for you to have a glance at them. Issues #131, #132, #133.

SSListItemFormat.java
SSDbArrayModel.java

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/models/SSDbArrayModel.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/models/SSDbArrayModel.java
@@ -87,7 +87,11 @@ public class SSDbArrayModel extends SSAbstractCollectionModel {
 	/** {@inheritDoc } */
 	@Override
 	public Object[] readData(final RowSet _rowSet, final String _columnName) throws SQLException {
-		return toObjList(getJDBCType(), _rowSet.getArray(_columnName)).toArray();
+		List<Object> data = toObjList(getJDBCType(), _rowSet.getArray(_columnName));
+		if(data == null) {
+			return null;
+		}
+		return data.toArray();
 	}
 
 	/** {@inheritDoc } */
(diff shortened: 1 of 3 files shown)


errael:
> have a glance at them

I looked yesterday when they were first pushed. Looked good to me.

swingset/src/main/java/com/nqadmin/swingset/models/SSDbArrayModel.java
```diff
@@ -87,7 +87,11 @@ public SSDbArrayModel(final JDBCType _jdbcType) {
 	/** {@inheritDoc } */
 	@Override
 	public Object[] readData(final RowSet _rowSet, final String _columnName) throws SQLException {
-		return toObjList(getJDBCType(), _rowSet.getArray(_columnName)).toArray();
+		List<Object> data = toObjList(getJDBCType(), _rowSet.getArray(_columnName));
+		if(data == null) {
+			return null;
```
errael, line 92: toObjList starts with a check an may return null. So definitely a bug that needed to be fixed here. I didn't look at the callers of readData to see if returning an empty array would be better.

swingset/src/main/java/com/nqadmin/swingset/models/SSListItemFormat.java
```diff
@@ -241,10 +241,10 @@ protected void appendValue(StringBuffer _sb, int _elemIndex, ListItem0 _listItem
...
 		case TIMESTAMP:
 			String pattern = patterns.get(jdbcType);
-			if (pattern != null) {
+			if (pattern != null && elem != null) {
 				SimpleDateFormat dateFormat = new SimpleDateFormat(pattern);
 				dateFormat.format(elem, _sb, FP0);
-			} else {
+			} else if(elem != null){
 				_sb.append(elem.toString());
 			}
```
errael, line 249: Definitely needs a fix. Taking a cue from the `default:` I might have wrapped this whole case in a single 
```
if(elem != null) {
    ...
}
```

PR #137, opened 2022-05-05
Component null pointer fixes

prasanthreddy-git:
Made another change to the createNullItem which is used to create a Item for the SSDBComboBox (used for Navigation) when a new record is added. Here were are using a empty string for the display value, when the field is a date field we try to format it and we added a null check but since we are using a empty string when creating a null item it causes a IllegalArgumentException. So changed the empty string to null when creating a NullItem

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
@@ -363,14 +363,25 @@ public class SSDBComboBox extends SSBaseComboBox<Long, Object, Object>
 	 * @param _option text that should be displayed in the combobox
 	 * @param _mapping  mapping of _option, typically a primary key
 	 */
-	public void addOption(String _option, Long _mapping) {
+	public void addOption(Object _option, Long _mapping) {
+		addOption(_option, null, _mapping);
+	}
+	
+	/**
+	 * Adds an item to the existing list of items in the combo box.
+	 *
+	 * @param _option text that should be displayed in the combobox
+	 * @param _option2  second display text for combobox
+	 * @param _mapping  mapping of _option, typically a primary key
+	 */
+	public void addOption(Object _option, Object _option2, Long _mapping) {
 		try (Model.Remodel remodel = optionModel.getRemodel()) {
 			final int index = remodel.getMappings().indexOf(_mapping);
 			if (index >= 0) {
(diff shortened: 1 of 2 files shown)


swingset/src/main/java/com/nqadmin/swingset/SSDBComboBox.java
```diff
@@ -196,7 +196,7 @@ private static class GlazedModel extends SSBaseComboBox.BaseGlazedModel<Long, Ob
 	 */
 	@Override
 	protected SSListItem createNullItem(Model.Remodel remodel) {
-		return remodel.createOptionMappingItem(null, "", null);
+		return remodel.createOptionMappingItem(null, null, null);
```
errael, line 199: Looks good. Nulls and Dates, who would've thought...

BTW, prompted me to take another look at `SSListItemFormat.appendValue(). Looks like `SSListItemFormat.setPattern()` is _only meaningful for date-time_ types. Sigh, now I'll obsess about it. One idea is that `setPattern()` should be deprecated in favor of `setFormatAndPattern()` then it could be readily used to format any jdbc data type.

Any objections if I take a look at that? If toString() is good enough for now and there's no short term need, then maybe I should do that to the formatting and masks branch.

errael (approved):

prasanthreddy-git:
What are the other database types for which we can automatically set some formatting? For varchar types, if the data is saved with the pattern (Ex SSN saved as xxx-xx-xxxx)  then there would not be a 
need to set pattern. But if only the row data is saved without the pattern then using the pattern would help.

Guess if the number saved is a dollar amount then we could format it and show that, if a pattern is provided.

errael:
Currently only `JDBCType.DATE`, `JDBCType.TIME`, `JDBCType.TIMESTAMP` can have patterns assigned. I'll put together a PR that allows everything to have a pattern.

prasanthreddy-git:
To what data types would the pattern apply?

errael:
I opened #138. With that, every data type can have it's own default Format (patterns are deprecated).

In addition, every element can have it's own Format which overrides the type's default.

errael:
> To what data types would the pattern apply?

I may have misunderstood this when I first read it. I guess there's an assumption that if an SSListItem element is tagged, for example as as JDBCType.DATE, then all the elements are of type. Or I guess something compatible.

Looks like SimpleDateFormat.format handles Date and Number.

PR #138, opened 2022-05-06
Use Format for SSListItemFormat; all types

errael:
This PR adds these features
- Each JDBCType can have a default Format
- Each SSListItem element can have Format

The use of a format string is deprecated (it still works, now built on the new Format mechanisms). Previously only the data/time types could have a format string; everything else was toString().

Here's some examples from the unit tests 
```src/test/java/com/nqadmin/swingset/models/SSListItemFormatTest.java```

Specifying defaults for a given jdbc type.
```
fmt.setFormat(JDBCType.INTEGER, new DecimalFormat("$#0000.00"));
fmt.setFormat(JDBCType.FLOAT, new DecimalFormat(".00"));
fmt.setFormat(JDBCType.DATE,      new SimpleDateFormat("YY"));
fmt.setFormat(JDBCType.TIMESTAMP, new SimpleDateFormat("YY*HH"));
...
expect = "23*07 @@@ 3.14 @@@ $0042.00 @@@ 21";

```

And overriding the defaults for SSListItem elements 0 and 1.
```
// add element format to override the default
fmt.clear();
fmt.addElemType(4, JDBCType.TIMESTAMP);
fmt.addElemType(1, JDBCType.FLOAT, new DecimalFormat(".0000"));
fmt.addElemType(0, JDBCType.INTEGER, new DecimalFormat("$.00"));
fmt.addElemType(2, JDBCType.DATE);
format = fmt.format(listItem);
expect = "23*07 @@@ 3.1416 @@@ $42.00 @@@ 21";
```

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/models/SSListItemFormat.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/models/SSListItemFormat.java
@@ -47,7 +47,10 @@ import java.util.EnumMap;
 import java.util.List;
 import java.util.Objects;
 
+import org.apache.logging.log4j.Logger;
+
 import com.nqadmin.swingset.models.AbstractComboBoxListSwingModel.ListItem0;
+import com.nqadmin.swingset.utils.SSUtils;
 
 // SSListItemFormat.java
 //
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by prasanthreddy-git** [comment]
Should we enclose all the code in appendValue in  a try catch so that IllegalArgumentException or incorrect formats don't break the system? We can just display text when formatting fails that would allow the screens to work even if formatting is not correct.

**Comment by errael** [comment]
Good catch. Looks like each subclass of Format might have it's own set of exceptions. My favorite is
```
ArithmeticException - if rounding is needed with rounding mode being set to RoundingMode.UNNECESSARY
```
Updated the PR.

Note that a bad Format with a long list generates lot of log entries.

I've added a couple of tests for different exception conditions. At some point I may know how to
test if a log entry was actually made, but until then...

_An aside._
I was wondering if at some point might want to have an optional Format for a null element.
If you run into something that might need that...

prasanthreddy-git:
Can we add ex.getMessage() to the error being logged, so that we know what exception is thrown?

errael:
> add ex.getMessage()

Will do. Had `ex` in there, felt like too much, can put it back. `ex.getMessage()` seems like a good compromise.

errael:
ERROR SSListItemFormat can't format true with java.text.SimpleDateFormat@50f478a. Exception: Cannot format given Object as a Date

prasanthreddy-git:
Thank you

prasanthreddy-git (approved):

Issue #145, opened 2022-06-08
SSComboBox/SSDBComboBox - Ability to display underlying database value when the combo doesn't have a mapping


prasanthreddy-git:
SSComboBox/SSDBComboBox show blank combo when there is no mapping available for the database value. Would be nice if we could show the underlying database value when there is no mapping string available.

errael:
I started with some confusion; the methods/comments refer to the displayed value as `option` and the primary key as `mapping`. In that context, I addressed displaying the `mapping` if there is no `option`. Let me know if I'm still misunderstanding.

prasanthreddy-git:
I think that is right. When we don't have a text to display we want to display what ever is in the database itself, in which it would be the mapping based on your comments.

PR #146, opened 2022-06-11
(Withdrawn) Show mapping if null option

errael:
Based on issue #145 
This is a draft PR, suggesting an approach to solving #145 . This is the first time `SSListItemFormat` is used as a base class; some difficulties were uncovered and are also addressed in this PR.

@prasanthreddy-git let me know how this approach handles your issue.


errael:
Ancient, droped in favor of #147 .

PR #147, opened 2022-06-12
Default combo SSListItemFormat shows Mapping if null Option.

errael:
Add `BaseComboBox.[sg]etListItemFormat` to improve display configurability.

This 2nd PR feels better for handling #145 . The externally visible change is the ability to set the list format on the combo rather than having an overriden method . So any given combo can more easily have it's display customized.

Internally, using a `SSListItemFormatDelegate` enables the improved configurability.

I was guessing that the display of mapping for null option was more for debug and/or identifying issues. Might want to change the default display.

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -468,7 +468,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 
 	/**
 	 * create SSBaseComboBox
-	 * @param useGlazed
+	 * @param useGlazed use glazedlists
 	 */
 	@SuppressWarnings("LeakingThisInConstructor")
 	public SSBaseComboBox(boolean useGlazed) {
@@ -839,7 +839,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 * from the database is encountered and no Option was specified,
 	 * a combobox list entry is created for the Mapping with null option;
 	 * a custom message, see
- 	 * {@link #setListItemFormat(com.nqadmin.swingset.models.SSListItemFormat),
+ 	 * {@link #setListItemFormat(com.nqadmin.swingset.models.SSListItemFormat)},
 	 * may be setup for handling these entries for the combobox display.
 	 * If {@code MOC_CLEANUP} then when the combobox navigates away
 	 * from a record with a missing option, the automatically created
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -347,7 +347,8 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 		@Override
 		protected void appendValue(StringBuffer _sb, int _elemIndex, SSListItem _listItem) {
 			if (getOptionFormatIndex() == _elemIndex
-					&& getElem(_elemIndex, _listItem) == null) {
+					&& getElem(_elemIndex, _listItem) == null
+					&& !Objects.equals(getNullItem(), _listItem)) {
 				Object key = getElem(getMappingFormatIndex(), _listItem);
 				_sb.append(key != null ? key.toString() : null)
 						.append(" - Option Not Found");
@@ -1116,6 +1117,14 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 * @return the created null item.
 	 */
 	protected abstract SSListItem createNullItem(BaseModel<M,O,O2>.Remodel remodel);
+
+	/**
+	 * Return this ComboBox's nullItem.
+	 * @return the nullItem
+	 */
+	public SSListItem getNullItem() {
+		return nullItem;
+	}
 
 	/**
 	 * A combobox used as a navigator has some restriction; for example,
--- a/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
@@ -49,6 +49,8 @@ import java.awt.event.KeyAdapter;
 import java.awt.event.KeyEvent;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 
@@ -62,6 +64,7 @@ import com.nqadmin.swingset.models.AbstractComboBoxListSwingModel;
 import com.nqadmin.swingset.models.GlazedListsOptionMappingInfo;
 import com.nqadmin.swingset.models.OptionMappingSwingModel;
 import com.nqadmin.swingset.models.SSListItem;
+import com.nqadmin.swingset.models.SSListItemFormat;
 import com.nqadmin.swingset.utils.SSCommon;
 import com.nqadmin.swingset.utils.SSComponentInterface;
 import com.nqadmin.swingset.utils.SSUtils;
@@ -201,6 +204,11 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 		{
 			//System.err.println(e.paramString());
 			if (e.getStateChange() == ItemEvent.SELECTED) {
+				// Move cleanupMissing... to updateSSComponent so that the
+				// combo list doesn't change while a row is displayed.
+				// If needed, could have a control option for when
+				// to do the cleanup.
+				// cleanupMissingMappingOptions();
 				if (getSelectedItem() != nullItem && selectionPending) {
 					setSelectionPending(false);
 					//System.err.println("itemStateChanged true --> false");
@@ -281,7 +289,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 		 */
 		protected static <M,O,O2>BaseGlazedModel<M,O,O2> install(SSBaseComboBox<M,O,O2> _jc) {
 			BaseGlazedModel<M,O,O2> model = new BaseGlazedModel<>();
-			model.autoComplete = AutoCompleteSupport.install(_jc, model.getEventList(), null, model.getListItemFormat());
+			model.autoComplete = AutoCompleteSupport.install(_jc, model.getEventList(), null, model.getListItemFormatDelegate());
 			
 			model.autoComplete.setFilterMode(TextMatcherEditor.CONTAINS);
 			model.autoComplete.setStrict(true);
@@ -303,6 +311,53 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	}
 
 	/**
+	 * @return the mapping index in an SSListItem
+	 */
+	public int getMappingFormatIndex() {
+		return optionModel.getMappingListItemElemIndex();
+	}
+
+	/**
+	 * @return the option index in an SSListItem
+	 */
+	public int getOptionFormatIndex() {
+		return optionModel.getOptionListItemElemIndex();
+	}
+
+	/**
+	 * @return the option2 index in an SSListItem
+	 */
+	public int getOption2FormatIndex() {
+		return optionModel.getOption2ListItemElemIndex();
+	}
+
+	/**
+	 * A list item formatter that displays the mapping if the option is null.
+	 */
+	@SuppressWarnings("serial")
+	class ShowMappingIfNullOption extends SSListItemFormat {
+
+		/**
+		 * If null option, then show the mapping;
+		 * otherwise do the default formatting.
+		 * @param _sb display value goes here
+		 * @param _elemIndex which option to format
+		 * @param _listItem combobox list item
+		 */
+		@Override
+		protected void appendValue(StringBuffer _sb, int _elemIndex, SSListItem _listItem) {
+			if (getOptionFormatIndex() == _elemIndex
+					&& getElem(_elemIndex, _listItem) == null) {
+				Object key = getElem(getMappingFormatIndex(), _listItem);
+				_sb.append(key != null ? key.toString() : null)
+						.append(" - Option Not Found");
+			} else {
+				super.appendValue(_sb, _elemIndex, _listItem);
+			}
+		}
+	}
+
+	/**
 	 * Log4j Logger for component
 	 */
 	private static final Logger logger = SSUtils.getLogger();
@@ -318,7 +373,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 * the nullItem must be set to null.</b>
 	 * @see #createNullItem(com.nqadmin.swingset.models.OptionMappingSwingModel.Remodel)
 	 */
-	protected SSListItem nullItem;
+	transient protected SSListItem nullItem;
 	
 //	/**
 //	 * String typed by user into combobox
@@ -358,7 +413,7 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	/**
 	 * The combo model.
 	 */
-	protected OptionMappingSwingModel<M,O,O2> optionModel;
+	transient protected OptionMappingSwingModel<M,O,O2> optionModel;
 
 	//////////////////////////////////////////////////////////////////////////
 	//
@@ -386,7 +441,9 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 * Convenience method for accessing the model with proper casting.
 	 * 
 	 * @return mapping list model with proper casting, may be null
+	 * @deprecated use field optionModel directly
 	 */
+	@Deprecated
 	protected OptionMappingSwingModel<M, O, O2> getOptionModel() {
 		return optionModel;
 	}
@@ -407,6 +464,49 @@ public abstract class SSBaseComboBox<M,O,O2> extends JComboBox<SSListItem> imple
 	 */
 	public SSBaseComboBox() {
 		addItemListener(new SSBaseComboBoxItemListener());
+	}
+
+	/**
+	 * create SSBaseComboBox
+	 * @param useGlazed
+	 */
+	@SuppressWarnings("LeakingThisInConstructor")
+	public SSBaseComboBox(boolean useGlazed) {
+		this();
+
+		if (useGlazed) {
+			optionModel = BaseGlazedModel.install(this);
+		} else {
+			optionModel = BaseModel.install(this);
+		}
+		optionModel.setListItemFormat(new ShowMappingIfNullOption());
+	}
+
+	/**
+	 * Set the format to use with this model.
+	 * <br>TODO: consider deferring the actual change until row change
+	 * 
+	 * @param _listItemFormat the format used with this model
+	 */
+	public void setListItemFormat(SSListItemFormat _listItemFormat) {
+		Object selectedItem = getSelectedItem();
+		boolean keep = Objects.equals(getEditor().getItem(), selectedItem);
+		optionModel.setListItemFormat(_listItemFormat);
+		// A new format may change the target contents of the comboBox editor.
+		// When it does change, the contents of the combo editor is changed
+		// from ListItem to a String with the old value.
+		// If the combo editor held the selected item before,
+		// then it should hold the selected item after.
+		if(keep)
+			getEditor().setItem(selectedItem);
+	}
+
+	/**
+	 * Return the listItemFormat associated with this combobox.
+	 * @return the associated listItemFormat
+	 */
+	public SSListItemFormat getListItemFormat() {
+		return optionModel.getListItemFormat();
 	}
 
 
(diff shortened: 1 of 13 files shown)


errael:
@prasanthreddy-git Will you have a chance to take a look at this? The new default list format shows how the list format can be overriden. And with `setListItemFormat` on the combobox, it should be relatively easy to have a custom display for specific combobox if/when required.

No rush, I just don't want it to get lost...

swingset/src/main/java/com/nqadmin/swingset/SSBaseComboBox.java
```diff
@@ -302,6 +303,25 @@ protected BaseGlazedModel() {
...
+	 */
+	@SuppressWarnings("serial")
+	class ShowMappingIfNullOption extends SSListItemFormat {
+
+		@Override
+		protected void appendValue(StringBuffer _sb, int _elemIndex, SSListItem _listItem) {
+			if (optionModel.getOptionListItemElemIndex() == _elemIndex
+					&& getElem(_elemIndex, _listItem) == null) {
+				_sb.append("=== ")
+					.append(getElem(optionModel.getMappingListItemElemIndex(), _listItem).toString())
```
errael, line 317: This is the exception line. Since optionModel has already been used, I think the only NPE would be trying to do toString() because of a null return from
```
getElem(optionModel.getMappingListItemElemIndex(), _listItem)
```
and, in OptionMappingSwingModel, there's
```
	public int getMappingListItemElemIndex() {
		return KEY_IDX;
	}
```
So, IIRC, this means that not only is the displayed value null, but the key is also null. This could be some startup or initialization condition, unless it's really a null key. Either way this looks like a reasonable solution:
```
Object key = getElem(optionModel.getMappingListItemElemIndex(), _listItem);
_sb.append("=== ")
	.append(key != null ? key.toString() : null)
	.append(" ===");

```
to avoid the NPE, and see if there's any further issue.

Since I don't see any tests as part of this PR, I guess I just used the existing tests and probably walked through this code. If this doesn't work, or you don't want to try the fix, provide a test case and I'll be happy to update the PR. Something added in the Demo would be good.

errael: BTW, if you run this with JDK-17 it will say exactly what caused the NPE

prasanthreddy-git: There is explicit key with value of null. But I do remember the combo boxes were updated to allow the user to remove a selection (empty selection which would set a null value in database). So the code probably is adding an option with null value as key to the list of user specified mappings.

errael: Yeah, that could be, maybe one of the demos will NPE. May want to special case null to leave it blank?

Guess that's for after it works...

prasanthreddy-git: using the below code so that we are showing empty item that users can select (to deselect a value). This is solving the NPE issue. Will test the display of key.

Object key = getElem(optionModel.getMappingListItemElemIndex(), _listItem); 
_sb.append(key != null ? key.toString() : "");

prasanthreddy-git: Tested with a combo which did not have the mapping needed for a key, the key was not displayed when the mapping is not present.

errael: Uh, too many negatives. Is this a good result? Sounds like it's a bad result.

In the code I think mapping means key, option means value.

prasanthreddy-git: Yeah, not getting the expected result. The combo box is blank when there is no option/value to display. Expecting to see the mapping/key from the database.

errael: Weird, since we're pretty sure it's getting there because of the NPE fix (but that could have been for a different record). That would seem to indicate that
```
Object key = getElem(optionModel.getMappingListItemElemIndex(), _listItem);
```
it returning a null.

I'm trying to see how I can reproduce this with the demo, the demo sql is at the following files, if you have any ideas on how I can reproduce the issue

https://github.com/bpangburn/swingset/blob/82abc676eebda08508c77ef5b539235f72aa8bbc/swingset-demo/src/main/resources/mysql.swingset-demo-app.sql#L45
https://github.com/bpangburn/swingset/blob/82abc676eebda08508c77ef5b539235f72aa8bbc/swingset-demo/src/main/resources/mysql.swingset-demo-components.sql#L44

I'll fire up the demo and see if I can see something

errael: Those are the wrong pointers; they are the mysql versions, The normal ones are in the same directory. I think I can do it, I'll let you know.

errael: On the line

https://github.com/bpangburn/swingset/blob/82abc676eebda08508c77ef5b539235f72aa8bbc/swingset-demo/src/main/resources/suppliers_and_parts.sql#L61

I change 'Jones' to NULL and I see a blank, I think that's the problem. I wonder if a NULL is getting converted to blank when the records are getting read from the database.

prasanthreddy-git: No. Sent the email via standard mail server.

errael: I had realized the 'Jones' thing was a red herring, when your mail came in.

With the change you suggested, I just saw two items (and the nullItem) in the combo box. So I tried
```
cmbPartColor.setOptions(new String[] { null, "Green", "Blue" });
```
And with that, here's what I see

NullOption

prasanthreddy-git: We need the ability to see the underlying database value without explicitly setting it in the options. If we know that we have to set, say code 0, we would set it with the right value. This request is to cover scenarios where 1. Developer forgot to add a key that is needed 2. Combo box items are based on some values in a database table but some how the value in the database (bound column) has some other value.

errael: That makes sense. I wonder where the record is being tossed out or ignored? 

The change you gave me
```
cmbPartColor.setOptions(new String[] { "Green", "Blue" }, new int[] {1,2});
```
covers case 1 I guess. 

How to reproduce case 2? Looks like the test might cover case 2 as well.

errael: Is this different behavior that pre-4 swingset? What used to happen in swingset 2/3?

prasanthreddy-git: This is the same behaviour we had right from the beginning. Showing the underlying database value rather than blank would be an enhancement.

errael: Always good to know I didn't break anything... Guess my "make it behave the same" skills worked... Actually, this is probably in code I didn't really change.

I'm in the middle of something, should be able to track down the troublesome code over the next few days. Then we can discuss how to handle it. Maybe keep the mapping in the list and give the missing option a null value, then the new list item format code would kick in and display the mapping.

prasanthreddy-git: I am guessing this needs a change in updateSSComponent of SSBaseComboBox. Maybe java would not allow us to do what we really want. We should not have that key as an option to choose when user tries to select a value. Should be used for display only.

errael: > We should not have that key as an option to choose

Why not? It's already in the data base, but I see what you mean. 

If it's in the combo, you can choose it; I guess the old behavior avoided choosing it :-) We could consider a flag that's part of the listItem that means can't be chosen, but that all sounds pretty horrible.

Isn't there some event when the combo chooses a value? At that point could ignore the choice and/or throw up an alert/warning.

errael: (there was a shift in requirements, found some time)

In `SSBaseComboBox` I added
```
+       boolean ADDITEM_HACK = true;
+       private int addItemHack(BaseModel<M,O,O2>.Remodel remodel, M _mapping) {
+               int index = -1;
+               if(_mapping != null) {
+                       if(ADDITEM_HACK) {
+                               remodel.add(_mapping, null);
+                               index = remodel.getMappings().indexOf(_mapping);
+                       }
+               }
+               return index;
+       }
```
and in `SetSelectedMapping(M _mapping)` I added
```
                        SSListItem item;
                        if (index != -1) {
                                item = remodel.get(index);
+                       } else if((index = addItemHack(remodel, _mapping)) != -1) {
+                               item = remodel.get(index);
                        } else {
                                // nullItem is either special first list item
```
And with the change you suggested to show the problem, when ex4 comes up

NullOpt-fix

errael: So does this support the developer, and is OK for production? 

I might display `*** 0 *** REPORT ISSUE` or some such.

Might want a property like `enableDiscoveredMapping` or something.

errael: In Example4, to verify that a custom list formatter could be done by the application, I've got
```
// SETUP THE COMBO BOX OPTIONS TO BE DISPLAYED AND THEIR CORRESPONDING VALUES
cmbPartColor.setListItemFormat(new SSListItemFormat() {
    @Override
    protected void appendValue(StringBuffer _sb, int _elemIndex, SSListItem _listItem) {
        if (cmbPartColor.getOptionFormatIndex() == _elemIndex
                && getElem(_elemIndex, _listItem) == null) {
            Object key = getElem(cmbPartColor.getMappingFormatIndex(), _listItem);
            _sb.append("*** ")
                    .append(key != null ? key.toString() : null)
                    .append(" *** REPORT ME");
        } else {
            super.appendValue(_sb, _elemIndex, _listItem);
        }
    }
    
});
//cmbPartColor.setOptions(new String[] { "Red", "Green", "Blue" });
cmbPartColor.setOptions(new String[] { "Green", "Blue" }, new int[] {1,2});
```
giving

NullOpt-fix2

errael:
I'll push an updated PR with the current state, I'll leave in the `ADDITEM_HACK` flag (without it there's no change in behavior. We can discuss what to do about it: forget it, make it an option, handle it all the time.

This change does satisfy the requirement of letting that developer know that there's a problem, and the developer can choose what to display if there's a missing option for the mapping.

errael:
This PR comes down to the changes made in Example4 to demo this new feature.

@bpangburn Please take a look when you get a chance to the change in `setSelectedMapping()`

bpangburn:
> This PR comes down to the changes made in Example4 to demo this new feature. The whole idea is to highlight a missing option, something shows up in the combo to show that there is a mapping from the database with no option specified.
> 
> @bpangburn Please take a look when you get a chance to the change in `setSelectedMapping()`

@errael , do you have a sense of what would be involved in stripping the missing mapping from the list after navigating away from the record with the offending value?

For Ex4 as written, the `*** 0 *** REPORT ME` remains in the list as you navigate to other records. I changed your test to `cmbPartColor.setOptions(new String[] { "Red","Blue" }, new int[] {0,2});` The first record/combo now looks fine as it finds 0==Red. When I go to the 2nd record then I get `*** 1 *** REPORT ME` in the combo and it persists when I navigate back/forth. Ideally, the REPORT ME items would only appear in the combo when there is an unmatched value in the database. I did not study all of the code above so no idea if this would be hard or easy, but if we're flagging when a REPORT ME item has been added to the list, then (back to my original question), maybe we can remove the last item from the combo when a navigation takes place?

errael:
To help me understand/clarify the intent/use of this feature...

It's my understanding that this situation only arises if there is a bug in the application. I had thought that during development this would always be enabled (only for SSComboBox of course). And it might be enabled in a release and rarely seen. The whole idea is that it helps to find problems. Out of curiosity, if today it was like this, how often would the app with a missing option escape into the wild?

Your request implies that you do not want a user to select the unnamed option. I don't entirely understand that; wasn't the option was left out by accident. So with what you're suggesting, if the current selection has a missing option it's OK to select an option with a name but not the other way around.

Anyway, I'll poke around in SSComboBox. I'm wondering this shouldn't be handled at a higher level, possibly with an assist from SSComboBox.

errael:
> what would be involved in stripping the missing mapping from the list after navigating away

add some code to do this

prasanthreddy-git:
This is not necessarily a coding issue. This can happen when the values in a combo box is based on user selection some where else. In that case if the user changes the original selection (in a 
different screen) then the value selected in the combo box is no longer valid. That that point either the user has to change the value selected in the combo box or undo there change to what are valid 
values for combo box.

In the swingset examples lets say the colors of the parts are selected by the user in a screen and then we display the color drop down based on valid colors user has selected in the Colors Screen. Now 
user has added a part and selected Red as the part color. Later on they go to the Colors Screen and remove red from the list of colors the parts can have. Now the color red they picked is no longer 
valid but we have to be able to show that the color of a part is set to Red so that they can change the color of that part to something else.

errael:
> > what would be involved in stripping the missing mapping from the list after navigating away
> 
> add some code to do this

My comment, done in haste, is misleading. I should have said

_I added some code to do this_. 

You can check it out. This whole patch is experimental, consider what to keep and how to enable any new behavior.

errael:
>

errael:
> This is not necessarily a coding issue. This can happen when the values in a combo box is based on user selection some where else. In that case if the user changes the original selection (in a different screen) then the value selected in the combo box is no longer valid. That that point either the user has to change the value selected in the combo box or undo there change to what are valid values for combo box. 

I'm not sure I agree that the above example is not a coding issue; and it may not be one you wish to solve. The underlying problem seems to be that screen/combo aren't alerted to changes in things they care about. I don't know databases that well, but IIUC you can't listen to changes, unless maybe you use some kind of stored procedure thing; anyway, I digress. Google's event bus offers a convenient way to broadcast/listen-to changes, but that's more digression. It might be interesting to consider how, in the distant future, things like this might be handled; at least providing an infrastructure so an app could keep things consistent within one JVM/app-instance.

Request for clarification: when you said "changes the original selection (in a different screen)". By screen do you mean different window on the same application/monitor? Or is this a different person/application changing something?

> In the swingset examples lets say the colors of the parts are selected by the user in a screen and then we display the color drop down based on valid colors user has selected in the Colors Screen. Now user has added a part and selected Red as the part color. Later on they go to the Colors Screen and remove red from the list of colors the parts can have. Now the color red they picked is no longer valid but we have to be able to show that the color of a part is set to Red

When you say
> show that the color of a part is set to Red

do you mean show a number/data-base-value that used to correspond Red?

> so that they can change the color of that part to something else.

Actually, I'm confused. Mappings are numbers that show up in the database, and Options are specified to the combo box by the application programmer. This is the situation that we've been considering. I can see how a programmer might take Mapping-Option values from some table an pass them into a combo box. But, IIRC, the colors map database table is not known to the combobox, so I'm not sure this is a good example.

prasanthreddy-git:
For this example assume that the colors are not hard coded but are provided by user in a different screen in the same application. See responses in blue below


On 10/31/22 10:32 AM, errael wrote:
>
>     This is not necessarily a coding issue. This can happen when the values in a combo box is based on user selection some where else. In that case if the user changes the original selection (in a
>     different screen) then the value selected in the combo box is no longer valid. That that point either the user has to change the value selected in the combo box or undo there change to what are
>     valid values for combo box.
>
> I'm not sure I agree that the above example is not a coding issue; and it may not be one you wish to solve. The underlying problem seems to be that screen/combo aren't alerted to changes in things 
> they care about. I don't know databases that well, but IIUC you can't listen to changes, unless maybe you use some kind of stored procedure thing; anyway, I digress. Google's event bus offers a 
> convenient way to broadcast/listen-to changes, but that's more digression. It might be interesting to consider how, in the distant future, things like this might be handled; at least providing an 
> infrastructure so an app could keep things consistent within one JVM/app-instance.
>
> Request for clarification: when you said "changes the original selection (in a different screen)". By screen do you mean different window on the same application/monitor? Or is this a different 
> person/application changing something? Yeah different window in the same application. Could be done by the same user or a different user.
>
>     In the swingset examples lets say the colors of the parts are selected by the user in a screen and then we display the color drop down based on valid colors user has selected in the Colors
>     Screen. Now user has added a part and selected Red as the part color. Later on they go to the Colors Screen and remove red from the list of colors the parts can have. Now the color red they
>     picked is no longer valid but we have to be able to show that the color of a part is set to Red
>
> When you say
>
>     show that the color of a part is set to Red
>
> do you mean show a number/data-base-value that used to correspond Red? Yes show the database value of Red in the drop down
>
>     so that they can change the color of that part to something else.
>
> Actually, I'm confused. Mappings are numbers that show up in the database, and Options are specified to the combo box by the application programmer. This is the situation that we've been 
> considering. I can see how a programmer might take Mapping-Option values from some table an pass them into a combo box. But, IIRC, the colors map database table is not known to the combobox, so I'm 
> not sure this is a good example.
>
I wish we have not used the term mappings to refer to the value stored in the database. When ever I think of mapping my mind goes to key-value pair.

errael:
> I wish we have not used the term mappings to refer to the value stored in the database. When ever I think of mapping my mind goes to key-value pair.

Can't say I disagree. IIRC, previously there was inconsistent naming, with `mapping` the most common for the database key. Could have another round of deprecation and renaming. How about `MapKey` or ???. What do you think of `Option`

errael:
> For this example assume that the colors are not hard coded but are provided by user in a different screen in the same application. 

So this is theoretical and has nothing to do with this PR?

> See responses in blue below

Unfortunately, I'm not seeing any colors. And I'm not sure your comments got through.

errael:
Oops, I seem to have lost/misplaced one of my comments.

Starting with #145, and taking the comment "stripping the missing mapping from the list after navigating away", this PR seems to cover the bases.

That doesn't mean considerations haven't evolved, and that this PR is not appropriate or desired anymore.

prasanthreddy-git:
I have forwarded the email to your email address so that the coloring is not lost.

errael:
Got it, thanks.

About left/right angle brackets. They seem to be taken out, along with the words between them. I still get caught by that.

errael:
> .
>> When you say
>>> show that the color of a part is set to Red
> 
>> do you mean show a number/data-base-value that used to correspond Red?
> 
> Yes show the database value of Red in the drop down
>>>so that they can change the color of that part to something else.

If a database value does not match something in the option list (supplied by the programmer when the combobox is set up and what is displayed in the combobox) is missing, then the current PR shows the  MappingKey from the database. The programmer can provide their own message, using a custom `SSListItemFormat` as shown in example 4.

The implementation does this by creating a "dummy" combo list entry with a null option value. This dummy entry is short lived by "stripping the missing mapping from the list after navigating away" as suggested.

This seems to satisfy #145 and the _short lived while selected_ request.

There may be other related issues; like how the Options (display values) are derived. Maybe starting a discussion about them would be good.

errael:
Added `setMissingMappingControl(EnumSet\<MissingMappingControl\> flags)` so programmer can control behavior when a mapping is encountered that has no specified option; some javadoc cleanup.

Other than possibly changing the defaults, should be ready to go. 

Whether your app wants to use it is another issue. This PR plugs a hole.

errael:
Doh! What's missing is Option, not Mapping. Change the names to reflect that, for example `MissingOptionControl`. This last PR update is only names, no code.

errael:
Arg. And correct a misspelling: opiton to option.

errael:
Oh crap. Look like I've added some stuff to this PR that doesn't belong. What's particularly weird is that it mentions something I just did, but says it happened two days ago. I'll try to get it fixed this evening. 

Would appreciate either comments, decline or merge.

errael:
I think it's OK now.

prasanthreddy-git:
I will test it with our app next week.

bpangburn:
This seems to be working well. I was a little concerned with all of the code added to Example4, but you were just providing an example of how to override the missing option control/display. Default is to show the value surrounded by "===" so "=== 0 ===" if we remove Red as an option using:
```cmbPartColor.setOptions(new String[] { "Green", "Blue" }, new int[] {1,2});```

I think a good default behavior would be to show one of the following (assuming a rowset value of 0 with no matching option):
0 - NO MATCH
0 - MISSING OPTION
0 - OPTION NOT FOUND

@prasanthreddy-git please weigh in.

errael:
> This seems to be working well. I was a little concerned with all of the code added to Example4, but you were just providing an example of how to override the missing option control/display. 

Right. Guess I thought this was easier that putting an example in javadoc, which should probably go on a todo list. But where to put it?

Speaking of where to put things, maybe these changes belong in `SSComboBox` rather than `SSBaseComboBox`. The downside is that it would require fiddling since the changes must be called from `SSBaseComboBox`. Hmm, it's not that hard, maybe should be moved.

The javadoc for overriding the default message could go in `SSComboBox`, since that's where it is most meaningful.

>Default is to show the value surrounded by "===" so "=== 0 ===" if we remove Red as an option using: `cmbPartColor.setOptions(new String[] { "Green", "Blue" }, new int[] {1,2});`
> 
> I think a good default behavior would be to show one of the following (assuming a rowset value of 0 with no matching option): 0 - NO MATCH 0 - MISSING OPTION 0 - OPTION NOT FOUND

Now that I finally understand that it is actually a missing option, I can change the default. I'm torn between

1. 0 - NO MATCH
2. 0 - MISSING OPTION
3. 0 - OPTION NOT FOUND

Would someone think that "0" is the missing option with 2.? Hmm, 3. seems the least chance of a misunderstanding.  (but **rebase first** to `SNAPSHOT2`). 

> 
> @prasanthreddy-git please weigh in.

prasanthreddy-git:
0 - Option Not Found sounds good to me.

The same logic would be needed for SSDBComboBox also. If you move the code to SSComboBox would you add it to SSDBComboBox also? Seems like having it in SSBaseComboBox would make the functionality 
available in both (SSComboBox and SSDBComboBox).

bpangburn:
@errael Any merit in putting a sample override and other similar things in the GitHub project wiki?

Guess another options would be to have a separate Example screen based on Example4 that just demonstrates overriding the behavior. My concern with having it in Example 4 is confusing someone trying to learn the basics. Same goes for the hotkey/mnemonic code that we added.

errael:
> in the GitHub project wiki?

That's a good idea. I guess the wiki hasn't been a first class citizen in my mind. Is it ready for this kind of stuff?

> concern with having it in Example 4 is confusing

Could have an `Example4a` or `Example4Advanced` or comments in `Example4` specifying what is advanced and can be ignored.

errael:
> The same logic would be needed for SSDBComboBox also.

Oh, OK. I was just guessing, I hadn't looked at that file.  So nothing to do in that regard.

Guess I was assuming that missing options were due to programming supplied list or something; I need to revisit `SSDBComboBox`.

bpangburn:
> > in the GitHub project wiki?
> 
> That's a good idea. I guess the wiki hasn't been a first class citizen in my mind. Is it ready for this kind of stuff?

Not sure how the wiki is perceived. I use the wiki for my release process documentation. Works well enough.
> 
> > concern with having it in Example 4 is confusing
> 
> Could have an `Example4a` or `Example4Advanced` or comments in `Example4` specifying what is advanced and can be ignored.

This is probably the better approach and less likely to get stale. I wonder if we could strip any advanced stuff out of Example4 and then extend Example4 as Example4Advanced with just the missing option/mapping customization, hotkey/mnemonic, and any other future 'master class' stuff.

errael:
Default missing mapping message changed to `# - Option Not Found`

Introduced Example4Advanced. A small wart, had to put `void addCustomListItemFormat() {};` in Example4 called _at just the right time_. I'm not sure, but I think something in there needs to happen before `bind()` and/or `SSSyncManager` init.

errael:
Latest push adds event notification when setListItemFormat; it doesn't seem to change anything.

Adds some test code to Example4Advanced to help track down curious behavior of setListItemFormat after things have been initialized. The test code is disabled in the push; enable it by FALSE --> TRUE.

errael:
Found the magic incantation to get this to work cleanly. Fun to run the test mentioned, and watch the displayed value change right away.

errael:
Note: the latest change only allows the `MissingOption` being removed in `updateSSComponent`. May want it to change as early as possible.

errael:
Fix: do `cleanupMissingMappingOptions()` at the end of updateSSComponent, after the next row's data is put in place.

errael:
Dropping into a nap, had a thought about sequential missing options of _different_ mappings. And yeah, a missing option for `3` followed by missing for `0` looked like
extra-missing

Original code was sensitive to where `cleanupMissingMappingOptions()` is called from. 

Simple fix, always check the list (don't exit early), and don't remove a mapping for the current selection.

bpangburn (commented):
Made one comment about the empty constructor for SSDBComboBox. Not particular to this PR, but wanted to see if it jogged anything in your memory.

Sending some nit-picky warnings from Eclipse by email. I can probably fix most of them if you want, but thought you should glance at them first.

PR #149, opened 2022-11-11
Fix SSTableModel draw after deleteRow.

errael:
Was playing around to check out `DataGrid`, ran into a drawing bug in `SSTableModel`. Simple one line fix.
Based change on `4.0.11.SNAPSHOT2` (just guessing).
Update Example5 to handle insert/delete row.


errael:
Just noticed that there are some debug print left, I'll get rid of those and push a change later today.

bpangburn:
This seems fine and doesn't appear to break <Ctl-x>, which is what we've always used for deletion. I don't know where that originated. @prasanthreddy-git Let me know if you're OK with the one line change to SSTableModel. I don't think it would have any unintended side-effects.

errael:
Notice that this change also adds some debugging buttons to Example 5. You can see the bug in action by selecting a row and hitting the delete button (with the one line change to SSTableModel commented out)

PR #150, opened 2022-11-12
Use environment variable, JAVA_PREFERRED_SCREEN, to place Windows.

errael:
No effect if env variable is not present. Could enhance Screens to also consider a property.

It's not hard to manually do this when working with SwingSet demos, but that easily leads to errors when pushing. This PR is not all that urgent, it is something I use a lot.

raelity-lib is pretty small, take a look at the source. You may find ValueMap of interest, extracts/maintains a map key extracted from the value.

### DIFF
--- a/swingset-demo/README.md
+++ b/swingset-demo/README.md
@@ -79,6 +79,9 @@ The swingset-demo uses an in-memory H2 database by default, but can be run using
 
 See "USING ALTERNATE DATABASE SERVERS" at the end of this document for more information.
 
+Note that the default screen for placement of the demo can be specified using the environment variable: `JAVA_PREFERRED_SCREEN`.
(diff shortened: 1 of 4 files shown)


bpangburn:
@errael  I didn't see the project listed under https://github.com/errael?tab=repositories. Are you hosting it elsewhere? I can also download the source from the Maven repository if needed. Possible I'm being really dense.

I don't mind adding a dependency much since it's for the demo and not the core swingset.

What are you setting as the value for JAVA_PREFERRED_SCREEN and what behavior is it triggering (or I can see if I know some values to try).

bpangburn:
Just to note it, the source for the Screens class is here: https://sourceforge.net/p/jvi/raelity-lib/ci/default/tree/lib/src/main/java/com/raelity/lib/ui/Screens.java

PR #158, opened 2023-01-10
Calling setDefaults in updateScreen so that default parent id gets

prasanthreddy-git:


swingset/src/main/java/com/nqadmin/swingset/utils/SSDataGridScreenHelper.java
```diff
@@ -539,6 +539,9 @@ public void updateScreen() {
 			
 			// UPDATE SELECTION CRITERIA FOR ANY OTHER SSDBCombos
 			updateSSDBComboBoxes();
+
```
bpangburn, line 542: Per the JavaDoc above this method, we indicate that setParentID() should be called prior to calling updateScreen(). I'm concerned that calling setDefaultValues() at the end of updateScreen() could overwrite values in a row unexpectedly.

prasanthreddy-git:
setDefaults of the DataGrid doesn't change values for existing record. These values are saved for use when a new record is added. This is unlike how setDefaults of a form view helper works.

PR #161, opened 2023-02-06
4.0.11 snapshot2

bpangburn:
@errael - I think this is up to date and ready for release other than removal of -SNAPSHOT. I set a target date of Feb. 8th, but we can adjust.

This is the first time we've had the new enforcement in place where I don't appear to be able to accept my own pull request and require an external reviewer.

No pressure to review every change, but see if it compiles and if the demo appears to work OK for you.


bpangburn:
It does give me an option to bypass the protections, but not a bad idea to have a quick look.

swingset/CHANGELOG.txt
```diff
@@ -3,13 +3,34 @@ ChangeLog file for the SwingSet Open Toolkit for Java Swing.
...
+	5. Fix for SSTableModel draw after a deletion:
+	   https://github.com/bpangburn/swingset/issues/156
+	6. Fix for updating of default parent ID when using
+	   SSDataGridScreenHelper:
+	   https://github.com/bpangburn/swingset/issues/157
+	7. Addition of screen helper method to combine setParentID()
+	   and updateScreen():
+	   https://github.com/bpangburn/swingset/issues/159
+	8. Addition of Example4 Advanced to SwingSet demo to demonstrate
+	   #5 above.
```
errael, line 28: Actually, (I'll leave the wordsmithing to you)
```
#. Add Example4 Advanced. Moved code from Example4 that demos
   navigator actions: assigning actions to function keys, 
   assigning actions to buttons.

   Add demo code showing use of MissingOptionControl and
   custom ListItemFormat for null;/missing Options.

#. Add DataGridExampleSupport used by Example5.
   Provides more testing/examples of grid usage.
   Improved H2 handling so demo can now do row
   insertion/deletion into grid. Exposed 5 above.
```

errael (approved):
Interestingly, most of the changes are for the new feature around missing option; glad to see that hole plugged. Does make it harder to see the changes that actually required another release.

Built it, several things loaded from maven central. Ran a few demos. The only code changes since I last worked with it are things I wouldn't see...

I just went ahead with the approve; you can decide about the `CHANGELOG`.

Didn't pay much attention to the POMs; found the `CHANGLELOG-POMS` interesting (publish it and share the pain)

swingset/CHANGELOG.txt
bpangburn: Edited CHANGELOG to pickup your changes. Thanks. Hopefully good enough. Full speed ahead.

PR #169, opened 2023-12-21
Merge 4.0.13-SNAPSHOT into formatting-and-masks-WORKING

bpangburn:
Want to bring the formatting-and-masks-WORKING branch up to date with 4.0.13-SNAPSHOT. Looks like there will be some conflicts to resolve.  I'll take care of these or abort if this turns out to be a bad idea.


errael:
LMK if there's something you want me to look at.

It looks like this is mostly about support files and just accepting (or whatever git/github speak is for taking one and throwing away the other) 4.0.13 would be OK. I don't think there was any new stuff added for format/mask; uh, well, maybe somthing in swingset-demo/pom.xml.

PR #172, opened 2023-12-27
Clean up Eclipse warnings.

errael:
The key to getting rid of the "raw type" warning (and making things look cleaner) was
- `final public GridComboEditorComboBox getComponent()`
   override to return some non-parameterized.
- Also defined `private final class GridComboEditorComboBoxi::getGridSelItem()`
  not required but got rid of some ugly casts.

What I don't understand is why the warning was there at all. It had a `@SuppressWarnings("unchecked")`.
Guessing it's an Eclipse compiler issue. Maybe they don't support this one on an assignment statement.

### DIFF
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/DataGridExampleSupport.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/DataGridExampleSupport.java
@@ -49,6 +49,8 @@ import java.awt.GridLayout;
 import java.awt.event.ActionEvent;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import javax.sql.RowSet;
 import javax.swing.AbstractButton;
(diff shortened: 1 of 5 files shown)


swingset-demo/src/main/java/com/nqadmin/swingset/demo/DataGridExampleSupport.java
```diff
@@ -183,7 +183,7 @@ private void setAllowDelete(AbstractButton b) {
 				+ "</center></html>");
 	}
 
-	@SuppressWarnings("deprecation")
+	@SuppressWarnings({"deprecation", "UseOfSystemOutOrSystemErr"})
```
bpangburn: Do we want to suppress deprecation warnings or should we try to fix them? Maybe no good alternative?

swingset-demo/src/main/java/com/nqadmin/swingset/demo/DataGridExampleSupport.java
```diff
@@ -308,6 +308,7 @@ private void setupDebugButtons() {
 	boolean[] hidden = new boolean[1];
 	int indexOrName = 0;
 
+	@SuppressWarnings("UseOfSystemOutOrSystemErr")
 	void outputColInfo() {
 		for (TableColumn col : dataGrid.getColumnsList()) {
 			System.err.printf("id: %s, widths: max %d, min %d, pref %d, set %d,\n",
```
bpangburn, line 313: Do we want to output to Log4j2?

bpangburn (commented):
Few minor comments. After I hear back, I'll complete the merge (or you can).

errael:
Will handle System.err/out by deleting or converting to log4j after the grid branch stabilizes.

bpangburn (approved):
Looks good.

PR #174, opened 2023-12-29
grid-WORKING-bp-edits

### LINKED ISSUE #7: Add a "sorter" to the SSDataGrid
2005-03-10, Prasanth Pasala: https://sourceforge.net/p/swingset/feature-requests/21/

As suggested in this thread on the forums, we need to investigate adding a table sorter to the SSDataGrid:
https://sourceforge.net/forum/message.php?msg_id=3038707

Amit provided the following info on how to implement this:

Also, I am using a TableSorter given by the sun
JTable table = (JTable) viewport.getView();
//Encapsulate the table model by TableSorter to add sorting functionality
TableSorter sorter = new TableSorter(table.getModel());
table.setModel(sorter);
sorter.setTableHeader(table.getTableHeader());

Is the additional comma due to the TableSorter, the code for the table sorter can be found at
http://java.sun.com/docs/books/tutorial/uiswing/components/example-1dot4/TableSorter.java (dead link)

Possibly:
https://docs.oracle.com/javase/tutorial/uiswing/examples/components/TableSorterDemoProject/src/components/TableSorter.java
From:
https://docs.oracle.com/javase/tutorial/uiswing/components/table.html

### LINKED ISSUE #126: SSDataGrid ComboBox Issue with duplicate entries
In SwingSet Demo, Example 6. Part #3 and #4 are both "Screw"

If you open Example 7 and look at the logging, it appears to be impossible to select the 2nd instance of "Screw". If you give a unique name to Part #4, the problem goes away.

Per @prasanthreddy-git :

> If you have a same name for two items in a combo, I think you can't select the second one. This is an issue with JComboBox...Probably we should change that to SSComboBox.

Per @errael:

> Yeah, I've heard that Jcombo inside Jtable is notoriously problematic with event handling, needing multiple clicks is the one I remember, but I think there's more. I vaguely recall that glazed said they could handle that better...JCombo uses .equals, not ==. Workaround is to wrap the string in a class that doesn't override equals.

Sounds like the best fix is to try to replace JComboBox with SSComboBox.

bpangburn:
All Eclipse suggested edits other than a bunch of missing JavaDoc parameters we can deal with later.

Also includes turning off table sorting by default.

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/SSDataGrid.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSDataGrid.java
@@ -848,7 +848,7 @@ public class SSDataGrid extends JTable {
 	@Override
 	public void setModel(TableModel _dataModel) {
 		// TODO: Support setModel to change SSTableModel?
-		if(getModel() instanceof SSTableModel)
+		if(super.getModel() instanceof SSTableModel)
 			throw new IllegalArgumentException("Can not change SSTableModel");
 		super.setModel(_dataModel);
 	}
--- a/swingset/src/main/java/com/nqadmin/swingset/SSDataGrid.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSDataGrid.java
@@ -305,7 +305,6 @@ public class SSDataGrid extends JTable {
 		transient final Object[] items;
 		transient final Object[] underlyingValues;
 
-		/** {@inheritDoc} */
 		public ComboEditor(final Object[] _items, final Object[] _underlyingValues) {
 			super(new GridComboEditorComboBox());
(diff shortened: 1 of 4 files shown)


errael (approved):
I think I expected dozens of pages of something or another, that's a relief. Nice to see the correct `{@linkplain...`. 

Is `{@inheritDoc}`  not the right thing?

errael:
Added a couple of issues that can be closed with merge.

bpangburn:
> Is `{@inheritDoc}` not the right thing?

It's the right thing for an overridden method. Otherwise there's nothing to inherit.

errael:
> > Is `{@inheritDoc}` not the right thing?
> 
> It's the right thing for an overridden method. Otherwise there's nothing to inherit.

Doh!

PR #177, opened 2024-01-05
Add SSVersion

errael:
Automatically make the version from pom.xml available to the running program.
Versions can be compared.

### DIFF
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/MainClass.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/MainClass.java
@@ -71,6 +71,7 @@ import org.h2.tools.RunScript;
 
 import com.nqadmin.swingset.models.SSCollectionModel;
 import com.nqadmin.swingset.models.SSMysqlSetModel;
+import com.nqadmin.swingset.utils.SSVersion;
 import com.raelity.lib.ui.Screens;
 
 import gnu.getopt.Getopt;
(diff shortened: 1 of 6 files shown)


swingset/src/main/java/com/nqadmin/swingset/utils/SSVersion.java
```diff
@@ -0,0 +1,208 @@
...
+		for (int i = 0; i < versionSequence.size(); ++i) {
+			int diff = versionSequence.get(i) - o.versionSequence.get(i);
+			if (diff != 0) {
+				return diff;
+			}
+		}
+		if (isSnapshot == o.isSnapshot) {
+			return 0;
+		}
+		// If this is a snapshot, then it is "smaller" that the other.
```
bpangburn: "smaller" than... vs "smaller" that...

swingset/src/main/java/com/nqadmin/swingset/utils/SSVersion.java
```diff
@@ -0,0 +1,208 @@
...
+	public static SSVersion get(String sVer) {
+		return parse(sVer);
+	}
+
+	/** Version number components. */
+	private final List<Integer> versionSequence;
+	/** Is the version a snapshot? */
+	private final boolean isSnapshot;
+
+	private SSVersion(List<Integer> _versionSequence, boolean _isSnapshot) {
```
bpangburn, line 116: Know it's not required, but I try to add JavaDoc for protected/private methods just for the developer benefit.

swingset-demo/src/main/java/com/nqadmin/swingset/demo/MainClass.java
```diff
@@ -635,11 +636,15 @@ private static void usage() {
 	 * @param _args - optional command line arguments, which are ignored by this
 	 *              program
 	 */
+	@SuppressWarnings("UseOfSystemOutOrSystemErr")
 	public static void main(final String[] _args) {
```
bpangburn, line 640: I was going to ask about logger vs system.x.print, but in a main method I think sending feedback to the terminal makes sense.

swingset/pom.xml
```diff
@@ -173,6 +173,12 @@
 	</dependencies>
 
 	<build>
+		<resources>
```
bpangburn, line 176: So this modification to the pom is telling it to dynamically modify the swingsetVersion.properties during build and replace ${project.version} with the {version} variable value at the top of the pom? Cool.

bpangburn (commented):
Neat!

swingset-demo/src/main/java/com/nqadmin/swingset/demo/MainClass.java
errael: Note that outputting the JDK version info has been there for year(s).

swingset/pom.xml
errael: All the files (both of them) in the directory are filtered/copied to the jar. It didn't seem worth it to create another resource directory.

PR #178, opened 2024-01-06
Fix error recovery after errors during `SSTextField` edit

### LINKED ISSUE #175: Text field has wrong value after error while editing
To reproduce:
1. Modify demo schema supplier_data to add `not NULL`
    `status SMALLINT not NULL`
2. Run Example 1, click on 1st records status field, cursor after `20`, make sure entire field not selected
3. Enter backspace, see `2` in status field
4. Enter backspace, get dialog `Null values are not allowed for STATUS`, press `OK`
Observe status field is blank
5. Click Next Record, then Prev Record
Observe status field shows `2`, which has been committed to the database.

Note: after the error popup, pressing `undo` puts `20` in the text field.

This is unexpected. At a minimum, after the error dialog, `2` should be shown in the text field.
Other possibilities include highlighting the field as an error, disabling all buttons until there's a valid entry.

This is related to #15; that issue is primarily about creating a new record.

errael:
Fix #175
Return boolean from `updateColumnText` indicating if an error occurred.
Add `DocumentFilter` to capture text field value before edit.
Use `SSSQLNullException` instead of `NPE`.
Deprecate methods that change `SSCommon` or `ssComponent`.
Add some constraints for `supplier_data` table to make it easier to test error recovery.

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/SSTextField.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSTextField.java
@@ -47,6 +47,7 @@ import java.util.StringTokenizer;
 
 import javax.sql.RowSet;
 import javax.swing.JTextField;
+import javax.swing.text.Document;
 
 import org.apache.logging.log4j.Logger;
 
@@ -62,6 +63,7 @@ import com.nqadmin.swingset.utils.SSUtils;
  * SSTextField extends the JTextField. This class provides different masks like
  * date mask, SSN mask etc.
  */
+// TODO: get rid of "mask"/... and use formatted text fields.
 public class SSTextField extends JTextField implements SSComponentInterface {
 
 	// TODO Consider adding an InputVerifier to prevent component from losing focus.
@@ -209,27 +211,36 @@ public class SSTextField extends JTextField implements SSComponentInterface {
 
 	}
 
-	/**
-	 * Type of mask to be used for text field (default = none).
-	 */
-	protected int mask = 0;
+	private static final int ANY_ALIGNMENT = 0xffff;
 
-	/**
-	 * Default number of decimals to show for float, double, etc.
-	 */
-	protected int numberOfDecimalPlaces = 2;
+	private static final int DEFAULT_MASK = 0;
+	/** Type of mask to be used for text field (default = none). */
+	protected int mask = DEFAULT_MASK;
+
+	private static final int DEFAULT_NUMBER_OF_DECIMAL_PLACES = 2;
+	/** Default number of decimals to show for float, double, etc. */
+	protected int numberOfDecimalPlaces = DEFAULT_NUMBER_OF_DECIMAL_PLACES;
 
 	/**
 	 * Common fields shared across SwingSet components
 	 */
-	transient protected final SSCommon ssCommon = new SSCommon(this);
+	transient protected final SSCommon ssCommon;
 
 	/**
 	 * Constructs a new, empty text field.
 	 */
 	public SSTextField() {
-		// Note that call to parent default constructor is implicit.
-		// super();
+		this(null, DEFAULT_MASK);
+	}
+
+	/**
+	 * Constructs a new text field with the specified text and mask.
+	 *
+	 * @param _text the text to be displayed.
+	 * @param _mask the mask required for this textfield.
+	 */
+	public SSTextField(final String _text, final int _mask) {
+		this(_text, null, null, _mask, DEFAULT_NUMBER_OF_DECIMAL_PLACES, ANY_ALIGNMENT);
 	}
 
 	/**
@@ -238,7 +249,7 @@ public class SSTextField extends JTextField implements SSComponentInterface {
 	 * @param _mask the mask required for this textfield.
 	 */
 	public SSTextField(final int _mask) {
-		this("", _mask);
+		this( _mask, DEFAULT_NUMBER_OF_DECIMAL_PLACES);
 	}
 
 	/**
@@ -249,8 +260,7 @@ public class SSTextField extends JTextField implements SSComponentInterface {
 	 * @param _numberOfDecimalPlaces number of decimal places required
 	 */
 	public SSTextField(final int _mask, final int _numberOfDecimalPlaces) {
-		this(_mask);
-		numberOfDecimalPlaces = _numberOfDecimalPlaces;
+		this(_mask, _numberOfDecimalPlaces, ANY_ALIGNMENT);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


swingset/src/main/java/com/nqadmin/swingset/SSTextField.java
```diff
@@ -478,10 +494,13 @@ public void setNumberOfDecimalPlaces(final int _numberOfDecimalPlaces) {
 	 * Sets the SSCommon data member for the current Swingset Component.
 	 *
 	 * @param _ssCommon shared/common SwingSet component data and methods
+	 * @deprecated Creates inconsistent state, no replacement.
 	 */
+	// TODO: is this used? What about all the listeners and such?
```
bpangburn: I think SSComponentInterface() could be modified to remove setSSCommon(). I did a reference search in Eclipse and its not used from within SwingSet. I also don't see references in our software using SwingSet.

I don't see any reason ssCommon would need to be replaced once a SwingSet component is instantiated.

Best guess here is that when I created SSCommon and SSComponentInterface, I was creating setter/getter pairs to be thorough, but I agree that it's dangerous to call setSSCommon().

errael:
Split and rebased. The only change I want to make is disabling events in `SSCommon::changedUpdate()`
```
if (!ok) {
        // restore previous text value
        if(previousValue != null) {
                // TODO: should disable document events similar
                // to how row set events are handled.
                // can make that change if needed/requested/required.
                ((JTextComponent) getSSComponent()).setText(previousValue);
        }
}
```
And clean up some leftover TODOs...

swingset/src/main/java/com/nqadmin/swingset/SSTextField.java
```diff
@@ -287,21 +296,44 @@ public SSTextField(final int _mask, final int _numberOfDecimalPlaces, final int
...
-	 * Constructs a new text field with the specified text and mask.
-	 *
-	 * @param _text the text to be displayed.
-	 * @param _mask the mask required for this textfield.
-	 */
-	public SSTextField(final String _text, final int _mask) {
+	/** All the contructors feed through here */
+	private SSTextField(String _text, final RowSet _rowSet, final String _boundColumnName,
+	final int _mask, final int _numberOfDecimalPlaces, final int _align) {
 		super(_text);
```
bpangburn, line 306: I wonder if we need both ssCommon and constructingSSCommon.

Rather than this:
```
	private SSCommon constructingSSCommon() {
		if (constructingSSCommon == null) {
			constructingSSCommon = new SSCommon(this);
		}
		return constructingSSCommon;
	}
```

Could we use this:
```
	private SSCommon constructingSSCommon() {
                SSCommon result = ssCommon;

		if (ssCommon == null) {
			result = new SSCommon(this);
		}
		return result;
	}
```

swingset/src/main/java/com/nqadmin/swingset/SSTextField.java
```diff
@@ -287,21 +296,44 @@ public SSTextField(final int _mask, final int _numberOfDecimalPlaces, final int
...
 	}
 
-	/**
-	 * Constructs a new text field with the specified text and mask.
-	 *
-	 * @param _text the text to be displayed.
-	 * @param _mask the mask required for this textfield.
-	 */
-	public SSTextField(final String _text, final int _mask) {
+	/** All the contructors feed through here */
```
bpangburn: All the **constructors**...(misspelling)

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
```diff
@@ -157,9 +166,29 @@ public void changedUpdate(final DocumentEvent de) {
...
+					boolean ok = true;
 					try {
-						setBoundColumnText(((javax.swing.text.JTextComponent) getSSComponent()).getText());
+						ok = setBoundColumnText(((JTextComponent) getSSComponent()).getText());
 					} finally {
+						if (!ok) {
+							// restore previous text value
+							if(previousValue != null) {
+								if (ssComponentListenerAdded) {
+									// avoid generating events while restoring text
```
bpangburn, line 182: Should we manage `listenerNeedsRestoration` inline here or should we move a state tracking variable to addSSDocumentListener() and removeSSDocumentListener() similar to addSSComponentListener() and removeSSComponentListener()?

This way you can call addSSDocumentListener() and it will add the missing listener if listenerNeedsRestoration is true and then set it to false once complete.

bpangburn (commented):
Made a comment where we _might_ be able to eliminate a variable.

Found one misspelling around line 303 of SSTextField.

Tested a bit against PostgreSQL and seems to be working fine.

swingset/src/main/java/com/nqadmin/swingset/SSTextField.java
errael: `constructingSSCommon()` is used from `createDefaultModel()` which needs to use the same SSCommon that ends up in the variable ssCommon. 

The real issue is that `JTextField` calls `createDefaultModel()` in its constructor; and the `SSTextField` constructor does `super(_text)` so the document model, which depends on sscommon, has to be in place before `SSTextField` is constructed. 

An alternative approach would be to let `JTextField` use it's own document model, then in `SSTextField`'s constructor create the real model, based on sscommon, copy the text from the `JTextField`'s model into the real model, and set the new model into `SSTextField`. That would be more work (creating two models) and since I tend to overoptimize...

I added a comment, which _may_ clarify things at least a little bit.

swingset/src/main/java/com/nqadmin/swingset/utils/SSCommon.java
errael: I don't know if that works; I'm not saying it doesn't. I'm not sure exactly what you are suggesting.

Is it possible that `addSSDocumentListener()` would be called while `listenerNeedsRestoration` is true? If it is, then I'm just confusing myself. But I don't see how that could happen.

I'm still not sure, in detail, what you're suggesting. This is more your bailiwick, LMK.

errael: Added a commit, https://github.com/bpangburn/swingset/pull/178/commits/eac0fc1dc82a2644121f0aa4f0238449138b4ded, with a TODO for this.

errael:
Added https://github.com/bpangburn/swingset/pull/178/commits/eac0fc1dc82a2644121f0aa4f0238449138b4ded to address review comments.

bpangburn (approved):
Approved pending @errael thoughts on extending this fix for SSTextArea and SSFormattedTextField.

PR #179, opened 2024-01-08
Make ssCommon/ssComponent final. Remove their setters

errael:
See https://github.com/bpangburn/swingset/pull/178#discussion_r1444070322 for some discussion


bpangburn (commented):
Changes look fine, but I want to test a bit against some existing screens with PostgreSQL before merging.

PR #181, opened 2024-01-09
Make grid database type usage consistent with RowSetOps

errael:
@prasanthreddy-git @bpangburn 

About using JDBCType Enum

Goals/issues
- Consistent treatment of `JDBCTypes` between grid and `RowSetOps`.
- Handling of java objects, currently `RowSetOps` works only with strings. (I think)
- Follow `JDBC` spec (**noticed some non-compliance in `RowSetOps`**);
  mostly because `FLOAT` is `Double`, `REAL` is `Float`.

In `SSTableModel` there are several places where's there's a TODO comment about
using `JDBCType`. I think those comments were already there from way back,
before this Grid PR. 

This PR changes a simple case:
`getColumnClass` modified using `RowSetOps` methods. `findJavaTypeClass()` handles
more types that Grid was using, and some are handled differently:
- FLOAT returns Double.class
- NUMERIC/DECIMAL return BigDecimal.class
- More time related types handled
- More string types handled
- default is an exception instead of Object.class

Take a look, if it makes sense, other places can be modified to match. I'm
unclear about the type mapping and/or how it interacts with the Grid. An
additional issue, in `SSDataGrid` there working on an object and casting it to a
specific type; that's currently not in `RowSetOps`. In `RowSetOps` it's mostly
working with a string. Maybe working with objects needs to be put into
`RowSetOps`.

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
@@ -568,11 +568,15 @@ public class RowSetOps {
 
 // TODO: Probably a better way to handle date to timestamp conversion. Formatter?
 // 2022-05-31_BP: Probably best to cast to a LocalDateTime and return if that fails.
-			Timestamp timestampValue;
+			Timestamp timestampValue = null;
 			try {
-				if (_updatedValue.length() == 19 || _updatedValue.length() > 20) {
-					//System.err.println("TIMESTAMP update: " + _updatedValue);
+				if (_updatedValue.length() == 10) {
+					Date dateValue = SSCommon.getSQLDate(_updatedValue);
+					timestampValue = new Timestamp(dateValue.getTime());
+				} else if (_updatedValue.length() == 19 || _updatedValue.length() > 20) {
 					timestampValue = java.sql.Timestamp.valueOf(_updatedValue);
+				}
+				if (timestampValue != null) {
 					_rowSet.updateTimestamp(_columnName, timestampValue);
 				}
 			} catch(IllegalArgumentException ex) {
@@ -781,6 +785,48 @@ public class RowSetOps {
 	 * @throws SQLException  thrown if a database error is encountered
 	 */
 	public static void updateColumnObject(RowSet _rowSet, Object _value, int _columnIndex, JDBCType type) throws SQLException {
+		if (Boolean.TRUE)
+			updateColumnObject1(_rowSet, _value, _columnIndex);
+		else
+			updateColumnObject2(_rowSet, _value, _columnIndex, type);
+	}
+
+	/**
+	 * Update the Grid's RowSet at the specified column index with the given Object value.
+	 * RowSet. Operate on the current row.
+	 * <p>
+	 * When the user changes/edits the SSDataGrid cell this method propagates the
+	 * change to the RowSet. A separate call is required to flush/commit the change
+	 * to the database.
+	 *
+	 * @param _rowSet RowSet on which to operate
+	 * @param _value string to be type-converted as needed and updated in
+	 *                      underlying RowSet column
+	 * @param _columnIndex   index of the database column
+	 * @throws SQLException  thrown if a database error is encountered
+	 */
+	public static void updateColumnObject(RowSet _rowSet, Object _value, int _columnIndex) throws SQLException {
+		if (Boolean.TRUE)
+			updateColumnObject1(_rowSet, _value, _columnIndex);
+		else
+			updateColumnObject2(_rowSet, _value, _columnIndex, getJDBCColumnType(_rowSet, _columnIndex));
+	}
+
+	/**
+	 * Update the Grid's RowSet at the specified column index with the given Object value.
+	 * RowSet. Operate on the current row.
+	 * <p>
+	 * When the user changes/edits the SSDataGrid cell this method propagates the
+	 * change to the RowSet. A separate call is required to flush/commit the change
+	 * to the database.
+	 *
+	 * @param _rowSet RowSet on which to operate
+	 * @param _value string to be type-converted as needed and updated in
+	 *                      underlying RowSet column
+	 * @param _columnIndex   index of the database column
+	 * @throws SQLException  thrown if a database error is encountered
+	 */
+	public static void updateColumnObject1(RowSet _rowSet, Object _value, int _columnIndex) throws SQLException {
 		_rowSet.updateObject(_columnIndex, _value);
 	}
 
--- a/swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
@@ -45,7 +45,8 @@ import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Time;
 import java.sql.Timestamp;
-import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.OffsetTime;
 import java.util.Calendar;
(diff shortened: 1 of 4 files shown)


swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
```diff
@@ -191,12 +191,12 @@ public static String getColumnText(final RowSet _rowSet, final String _columnNam
...
 				break;
 
-			case FLOAT:
+			case REAL:
 				value = String.valueOf(_rowSet.getFloat(_columnName));
 				break;
 
 			case DOUBLE:
-			case REAL:
+			case FLOAT:
```
bpangburn, line 201: Seems really helpful that JDBCType of FLOAT is an alias for JDBCType of DOUBLE. No possible risk of crossing that up!
I think it would be good to add some comments in the code to point to the JDBC 4.2 spec. Specifically Appendix B starting on page 204.
https://download.oracle.com/otn-pub/jcp/jdbc-4_2-mrel2-spec/jdbc4.2-fr-spec.pdf

bpangburn (commented):
I agree that we need to clean up float/real/double as you have done. I suggested that we add the JDBC 4.2 spec and point to appendix B for quick reference. JDBC type of FLOAT being an alias of DOUBLE is fun.

errael:
In `RowSetOps` add method `updateColumnObject` which updates a `RowSet`
column with a value. There are actually two versions of this method,
one just simple does `rowSet.updateObject()` and lets the jdbc driver
do the conversion.

errael:
@bpangburn For `RowSetOps`, AFAICT, this is good for review. Includes fixes 
- better support `TIMESTAMP`
- `TIME` was added

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
```diff
@@ -669,9 +733,119 @@ public static Class<?> findJavaTypeClass(final JDBCType _jdbcType)
...
+	 *
+	 * @param _rowSet RowSet on which to operate
+	 * @param _value string to be type-converted as needed and updated in
+	 *                      underlying RowSet column
+	 * @param _columnIndex   index of the database column
+	 * @param type The JDBCType of the column
+	 * @throws SQLException  thrown if a database error is encountered
+	 * @see <a href="https://download.oracle.com/otn-pub/jcp/jdbc-4_3-mrel3-eval-spec/jdbc4.3-fr-spec.pdf">JDBC 4.3 Specification</a> Appendix B
+	 */
+	public static void updateColumnObject2(RowSet _rowSet, Object _value, int _columnIndex, JDBCType type) throws SQLException {
```
prasanthreddy-git, line 849: Can we get the JDBCType by calling the getJDBCColumnType function rather than getting JDBCType as argument?

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
```diff
@@ -519,36 +547,51 @@ public static void updateColumnText(final RowSet _rowSet, final String _updatedV
...
+			Timestamp timestampValue;
+			try {
+				if (_updatedValue.length() == 19 || _updatedValue.length() > 20) {
+					//System.err.println("TIMESTAMP update: " + _updatedValue);
+					timestampValue = java.sql.Timestamp.valueOf(_updatedValue);
+					_rowSet.updateTimestamp(_columnName, timestampValue);
+				}
+			} catch(IllegalArgumentException ex) {
+				logger.warn("updateColumnText: TIMESTAMP: " + ex.getMessage());
+				throw ex;
```
prasanthreddy-git, line 584: If only date is entered, string length 10, we should probably support that. Based on the above code database will not be updated if only date is entered.

errael: > If only date is entered, string length 10, we should probably support that. Based on the above code database will not be updated if only date is entered.

Why is that a problem?

But OK, I'll put it back.

Makes me wonder why a database column would be a TIMESTAMP and not a DATE if all you needed is date. I don't think TIMESTAMP or TIME is ever used; it looks like they would have gotten an exception on almost every keystroke.

Hopefully the mask formatting stuff can be finished and eventually find it's way into the grid parsing. It's weird that only one date format is accepted.

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
errael: I thought, but could be very wrong, that accessing meta data was very expensive.

I'll make the two changes discussed below.

I see that `updateColumnObject()` is used in 3 places. Could have an alternate version of `updateColumnObject()` that only takes two arguments which means that the type should be figured out automatically. That would be best because one version assumes doesn't need the type info.

Notice that `updateColumnObject()` and `updateColumnObject2()` take the same args. One depends on `updateObject` working, the other is a fallback; since the object type is correct it problably works, but the alternate is available. One thought is to have xxx1() and xxx2 both, and have `updateColumnObject()` calls one or the other, making it even simpler.

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
errael: DONE

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
errael: DONE

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
prasanthreddy-git: Agree that the column type should be Date in those cases but if the user types in just the date it would be good if we can update the database with that date rather than the update failing, especially because it have been working until now.

swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
prasanthreddy-git: I was more thinking of just having one function with below signature. Then the first line of this function would be JDBCType  type = RowSetOps.getJDBCColumnType(rowset, _column + 1);
The switch block would follow this line.

public static void updateColumnObject(RowSet _rowSet, Object _value, int _columnIndex) throws SQLException {

Line 850 of SSTableModel will be updated to remove the type argument. Which is the only place where we are passing type argument to updateColumnObject.

errael: I understand. I have a tendency to over-optimize. The function you mention
```
public static void updateColumnObject(
    RowSet _rowSet, Object _value, int _columnIndex) throws SQLException {
```
exists. The other funtion
```
public static void updateColumnObject(
    RowSet _rowSet, Object _value, int _columnIndex, JDBCType type) throws SQLException {
```
is available if you already have the type so you can avoid accessing meta data. (I thought that was expensive).

And in some sense, I see the stuff in RowSetOps in a transitional state, since I'm not sure what works with any give jdbc driver. I'm hoping that we can get rid of the type entirely since we've already got an object of the expected type. Notice that the way it's set up now
```
RowSetOps.getJDBCColumnType(rowset, _column + 1);
```
is never executed since
```
_rowSet.updateObject(_columnIndex, _value);
```
doesn't take a type.

prasanthreddy-git (approved):
Once everything is merged to gridReview branch and tested. We could probably get rid of multiple updateObject methods and have just one of them, if updates work fine without type, else have two functions one with type and another without type which intern calls the function with type argument.

PR #182, opened 2024-01-10
Changes after Grid review.

errael:
- Return null instead of -1 if, in `GridCombo`, no underlying values and nothing selected.
- Fix bug with doing `fireTableStructureChanged`, don't do it if not changed.
- Always confirm deletion after Control-X; use `SSDataGrid` if no message window specified.
- Clean up some comments.


prasanthreddy-git (commented):
In the else block do we need this assignment if they are the same?
               
if (colClasses.equals(columnClasses)) {
			return false;
		} else {
			columnClasses = colClasses;
			return true;
		}

errael:
> In the else block do we need this assignment if they are the same?
> 
> ```
> if (colClasses.equals(columnClasses)) {
>     return false;
> } else {
>     columnClasses = colClasses;
>     return true;
> }
> ```

The `return false` means `columnTypesChanged()` is false which means nothing changed

prasanthreddy-git (approved):

PR #185, opened 2024-01-16
SSFormattedTextField/SSComponent Decorating

errael:
Take the gui highlighting code out of SSFormattedTextField and use it in a new highlights plugin capability. The new highlight plugins are usable with any SSComponent type.

- `Hightlight` or `Decorate`? Which do you like better?
- How about the partitioning? A separate directory or a directory with a different name?
- Other issues?


errael:
Will delete ghost code from SSFormattedTextField before taking this PR out of draft.

errael:
Moved to 4.0.13

PR #186, opened 2024-01-18
SSComponent DecoratorValidator infrastructure; incorporated into SSFormattedTextField

errael:
Plugin architecture for decorating any SSComponent.
Current decorators all inherit from abstract FocusDecorator.java.
In SSFormattedTextField, use it with some customization.

### DIFF
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example1.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example1.java
@@ -54,6 +54,7 @@ import org.apache.logging.log4j.Logger;
 import com.nqadmin.swingset.SSDBNavImpl;
 import com.nqadmin.swingset.SSDataNavigator;
 import com.nqadmin.swingset.SSTextField;
+import com.nqadmin.swingset.decorators.TextComponentValidator;
 
 /**
  * This example displays data from the supplier_data table.
@@ -116,6 +117,14 @@ public class Example1 extends JFrame {
 			
 		// SET SCREEN POSITION
 			setLocation(DemoUtil.getChildScreenLocation(this.getName()));
+		
+		// SET A VALIDATOR (may be a no-op is disabled in SwingSet library)
+			txtSupplierName.getSSCommon().setValidator(new TextComponentValidator() {
+				@Override
+				public boolean validate() {
+					return !jc().getText().equalsIgnoreCase("oops");
+				}
+			});
 
 		// INITIALIZE DATABASE CONNECTION AND COMPONENTS
 			try {
--- a/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example1.java
+++ b/swingset-demo/src/main/java/com/nqadmin/swingset/demo/Example1.java
@@ -119,12 +119,14 @@ public class Example1 extends JFrame {
 			setLocation(DemoUtil.getChildScreenLocation(this.getName()));
 		
 		// SET A VALIDATOR (may be a no-op is disabled in SwingSet library)
-			txtSupplierName.getSSCommon().setValidator(new TextComponentValidator() {
-				@Override
-				public boolean validate() {
-					return !jc().getText().equalsIgnoreCase("oops");
-				}
-			});
+			txtSupplierName.getSSCommon().setValidator(TextComponentValidator.create(
+					(jtc) -> !jtc.getText().matches("(?i).*oops.{0,2}$")));
+			//txtSupplierName.getSSCommon().setValidator(new TextComponentValidator() {
+			//	@Override
+			//	public boolean validate() {
+			//		return !jc().getText().equalsIgnoreCase("oops");
+			//	}
+			//});
 
 		// INITIALIZE DATABASE CONNECTION AND COMPONENTS
 			try {
--- a/swingset/src/main/java/com/nqadmin/swingset/SSCheckBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSCheckBox.java
@@ -158,7 +158,7 @@ public class SSCheckBox extends JCheckBox implements SSComponentInterface {
 	/**
 	 * Common fields shared across SwingSet components
 	 */
-	transient protected final SSCommon ssCommon = new SSCommon(this);
+	transient protected final SSCommon ssCommon;
 
 	/**
 	 * Unchecked value for numeric columns.
@@ -169,11 +169,7 @@ public class SSCheckBox extends JCheckBox implements SSComponentInterface {
 	 * Creates an object of SSCheckBox.
 	 */
 	public SSCheckBox() {
-		// Note that call to parent default constructor is implicit.
-		//super();
-		setBorderPainted(true);
-		Insets i = getInsets();
-		setBorder(BorderFactory.createEmptyBorder(i.top, i.left, i.bottom, i.right));
+		this(null);
 	}
 	
 	/**
@@ -187,7 +183,7 @@ public class SSCheckBox extends JCheckBox implements SSComponentInterface {
 	 * @throws SQLException - if a database access error occurs
 	 */
 	public SSCheckBox(final RowSet _rowSet, final String _boundColumnName) throws java.sql.SQLException {
-		this();
+		this(null);
 		bind(_rowSet, _boundColumnName);
 	}
 
@@ -198,6 +194,10 @@ public class SSCheckBox extends JCheckBox implements SSComponentInterface {
 	 */
 	public SSCheckBox(final String _text) {
 		super(_text);
+		Insets i = getInsets();
+		setBorder(BorderFactory.createEmptyBorder(i.top, i.left, i.bottom, i.right));
+		setBorderPainted(true);
+		ssCommon = new SSCommon(this);
 	}
 
 	/**
--- a/swingset/src/main/java/com/nqadmin/swingset/SSCheckBox.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSCheckBox.java
@@ -46,10 +46,12 @@ import java.sql.SQLException;
 import javax.sql.RowSet;
 import javax.swing.BorderFactory;
 import javax.swing.JCheckBox;
+import javax.swing.border.Border;
+import javax.swing.border.CompoundBorder;
 
 import org.apache.logging.log4j.Logger;
 
-import com.nqadmin.swingset.decorators.Decorator;
+import com.nqadmin.swingset.decorators.BorderDecorator;
 import com.nqadmin.swingset.utils.SSCommon;
 import com.nqadmin.swingset.utils.SSComponentInterface;
 import com.nqadmin.swingset.utils.SSUtils;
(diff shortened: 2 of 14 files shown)


errael:
Updated this PR with a simple `Validator` infrastructure. The default `Decorator` invokes the `Validator`.

As a test, modified `Example1` adding:
```
txtSupplierName.getSSCommon().setValidator(new TextComponentValidator() {
    @Override
    public boolean validate() {
        return !jc().getText().equalsIgnoreCase("oops");
    }
});
```
Enter a supplier name of "Oops" and get error state.

bpangburn:
Here's what I found poking around:

1. SSComboBoxes appear to have an extra border around them, but the border does not turn green when that field has the focus.
2. In Test Base Components, the images are not rendering, but if I pick a new image, it displays OK. The image is rendered/loaded properly in 4.0.13-SNAPSHOT.  No green border highlighting.
3. In Test Base Components, the checkbox border does not turn green when that field has the focus.

I vaguely recall you mentioning something about images not loading properly, but I thought that was a CachedRowSet issue.

errael:
> Here's what I found poking around:

This is one of those "how did anything work" moments.
Take a look at `com.nqadmin.swingset.decorators.BorderDecorator`.

I think I copied stuff from FormattedTextField, stuff just worked (mostly) and I forgot about it. A key indication that work is need:
```
defaultBorder = UIManager.getBorder("FormattedTextField.border")
```

Some initial thoughts
- Glad that each component has it's own decorator (not some over optimization attempt to share).
- Why is there both standardBorder and defaultBorder?
- Capture the `standardBorder` from the JComponent during the first decorate and don't use the UIManager. (maybe caputure during install) Nuke the default border.
- Use the width/thickness of the captured border to generate the focusBorder and error border.
- Spend some time wondering why the RED line border is created every time.
- Refresh my memory of how borders work. Consider border creation might be cheap.
- BorderFactory?
- Add some logging to gather information about the non-JTextField components and their borders.

After that, consider what might be needed, experiment...

If you have any insights or suggestions, I'm listening...

>     3. In Test Base Components, the checkbox border does not turn green when that field has the focus.

Should it? Could be a special case: create a border where none exists, may need something like that for combo boxes. But in the comboBox case there could be inherent weird border issues.

> 
> I vaguely recall you mentioning something about images not loading properly, but I thought that was a CachedRowSet issue.

Right, a CachedRowSet issue.

bpangburn:
First, regarding the image display in Test Base Components, I don't think it's a CachedRowSet thing. If I open **4.0.13-SNAPSHOT**, the image appears fine for "SHARE_JDBC" and "NQADMIN" and for "POOL_CACHED" it shows the text "No Picture".

If I open Decorate, SSImage shows the following for all three RowSet options:
image

Second, nuking defaultBorder seems to make sense.

Third, I think if the green border is indicating the field with the focus then ideally we'd find a way to make it work for all components. May or may not be feasible:

1. SSSlider - works
2. SSImage - does not work; not sure if we can stick SSImage in a border?
3. SSLabel - does not work; since it can't get the focus, I think this is expected/fine
4. SSList - works
5. SSComboBox/SSDBComboBox - does not work; this is a compound(?) component so maybe there are multiple borders involved (e.g. one for list and one for text field); possible that we're not applying the decorator to the correct sub-component. 
6. SSCheckBox - does not work; maybe we can wrap it in a border so its appearance is similar to SSSlider?

Fourth, based on the following snippet from SSCommon.initDecorator(), I'm not sure if your intent was to only decorate SSFormattedTextFields. If I set `DISABLE_GENERAL_VALIDATION = true;` then it attempts to decorate most components.

```
	// Only decorate/validate for SSFormattedTextField, as in v4.0.12.
		if (!(getSSComponent() instanceof SSFormattedTextField)
				&& DISABLE_GENERAL_VALIDATION) {
			return;
		}
```

And to be clear, my ideal outcome would be to try and give any component with the focus a green border.

bpangburn:
@errael  Guess I didn't need to write up that comment as you were clearly reading my mind.

errael:
> First, regarding the image display in Test Base Components, I don't think it's a CachedRowSet thing. If I open **4.0.13-SNAPSHOT**, the image appears fine for "SHARE_JDBC" and "NQADMIN" and for "POOL_CACHED" it shows the text "No Picture".

Right. I put in a worth while feature and used it do disable the `SSImage` decoration as a workaround. Without the `SSComponentInterface.createDefaultDecoror()`, doing `setDecorator` in the `SSImage` constructor was too late to prevent the weirdness. Not sure what's going on, will have to investigate. I wonder, could it be because the button gets focus?

> Second, nuking defaultBorder seems to make sense.

Not exactly sure what you mean, for SSImage?

> Third, I think if the green border is indicating the field with the focus then ideally we'd find a way to make it work for all components. 

Agree.

> May or may not be feasible:

>     2. SSImage - does not work; not sure if we can stick SSImage in a border?

Comment is above

>     5. SSComboBox/SSDBComboBox - does not work; this is a compound(?) component
>        so maybe there are multiple borders involved (e.g. one for list and one
>        for text field); possible that we're not applying the decorator to the
>        correct sub-component.

It is weird, doing `combo.getBorder()` returns null. I haven't looked at the `JCombobox` code yet, there might be something to tweak which could be an alternate fix.

The way it's set up now, it puts a width 1 border around the `JCombobox` (because it has no border) that changes the layout a little.

Look OK?

>     6. SSCheckBox - does not work; maybe we can wrap it in a border so its appearance is similar to SSSlider?

Turns the the `JCheckBox` code turned off painting the button border. Fixed in `SSCheckbox` constructor with instruction to do the paint and replacing the border with an `EmptyBorder` of the same size `[4,4,4,4]`. The green border around the checkbox seems a little close top/bottom. Since it's such a wide border, thnking about putting a little space.

What do you think?

Hmm, something I did recently broke the "make an empty border" probably the `createDefaultDecorator()`; something to look at.

> Fourth, based on the following snippet from SSCommon.initDecorator(), I'm not sure if your intent was to only decorate SSFormattedTextFields. If I set `DISABLE_GENERAL_VALIDATION = true;` then it attempts to decorate most components.
> 
> ```
> 	// Only decorate/validate for SSFormattedTextField, as in v4.0.12.
> 		if (!(getSSComponent() instanceof SSFormattedTextField)
> 				&& DISABLE_GENERAL_VALIDATION) {
> 			return;
> 		}
> ```

Do you see a bug when you run things? The idea with the above, was that if the general validation is turned off, it is still needed for SSFormattedTextField. Looking at the code above, it doesn't look quite right, but I don't see a problem when I run the demo.

> And to be clear, my ideal outcome would be to try and give any component with the focus a green border.

Right, seems like `SSImage` is the only problem left; have to see what investigation shows up.

errael:
Think all the problems/issues are taken care of (except `SSImage` doesn't get decorated).

`SSImage` might have a UI issue since the button gets focused.

bpangburn:
For SSImage, I wonder what it would look like if we gave the button a green border rather than the image?

Only other thing I see above that I think you wanted me to clarify was:

> >    Second, nuking defaultBorder seems to make sense.
> >
> Not exactly sure what you mean, for SSImage?

This was in response to your comment from yesterday:

>>- Why is there both standardBorder and defaultBorder?
>>- Capture the standardBorder from the JComponent during the first decorate and don't use the UIManager. (maybe capture during install) **Nuke** the default border.

So I think your Nuke comment was related to a possible extraneous variable. I don't think it was related to SSImage.

I just pulled the latest and the Decorating is working great. Nice work!!

errael:
The latest commit, https://github.com/bpangburn/swingset/pull/186/commits/f63664a6f84ac1ab80cfce0c293a7bdebe1aa492, should finish this off.

There is a remaining issue in the component demo around the `SSList`. Notice that the border is on the list, it should be on the scroll pane I think. The border scrolls around.

Given the current `SSList` component, in `TestBaseComponents` (where the scroll pane is created) a custom decorator should be created that uses the scroll pane as where the border should go. Probably the correct solution is to add an option to `SSList` to add a scroll pane, if this option is taken then `SSList` should automatically adjust its default border decorator.

What's going on with `SSList` might give a clue to the source of the weird `SSImage` behavior before decorating its button.

errael:
Including the following is for history. It's what prompted doing the various methods for deriving a border for decoration.

ssimage-update-border

bpangburn:
> The latest commit, [f63664a](https://github.com/bpangburn/swingset/commit/f63664a6f84ac1ab80cfce0c293a7bdebe1aa492), should finish this off.
> 
> There is a remaining issue in the component demo around the `SSList`. Notice that the border is on the list, it should be on the scroll pane I think. The border scrolls around.
> 
> Given the current `SSList` component, in `TestBaseComponents` (where the scroll pane is created) a custom decorator should be created that uses the scroll pane as where the border should go. Probably the correct solution is to add an option to `SSList` to add a scroll pane, if this option is taken then `SSList` should automatically adjust its default border decorator.
> 
> What's going on with `SSList` might give a clue to the source of the weird `SSImage` behavior before decorating its button.

I think highlighting the button on an SSImage is OK. Would it be better to have a green frame around the whole image? Maybe, but there is some logic to highlighting the thing you can edit.

As for the SSList and scroll pane, I'm OK with the scroll bars not being framed in green. I don't feel strongly.

errael:
> I think highlighting the button on an SSImage is OK. 

I agree. And that's what gets focused, and the only place the user can do something. And it looks OK to me.

> Would it be better to have a green frame around the whole image? Maybe, but there is some logic to highlighting the thing you can edit.

I only mentioned `SSImage` because I never pinpointed why things screw up by setting a border; there's something lurking.

> As for the SSList and scroll pane, I'm OK with the scroll bars not being framed in green. I don't feel strongly.

It's a visual wart to have the green border scrolling, the colored border top and bottom scroll out of view; that's just plain not right. It points out that if someone is going to wrap an `SSComponent` in something like scroll pane it has responsibilities to insure that border highlighting works. Consider if you can scroll both horizontally and vertically, the information provided by the border highlight can be scrolled entirely out of view.

I wasn't suggesting any further work is needed at this time. Just some things to note and keep in mind...

bpangburn:
> > As for the SSList and scroll pane, I'm OK with the scroll bars not being framed in green. I don't feel strongly.
> 
> It's a visual wart to have the green border scrolling, the colored border top and bottom scroll out of view; that's just plain not right. It points out that if someone is going to wrap an `SSComponent` in something like scroll pane it has responsibilities to insure that border highlighting works. Consider if you can scroll both horizontally and vertically, the information provided by the border highlight can be scrolled entirely out of view.
> 
> I wasn't suggesting any further work is needed at this time. Just some things to note and keep in mind...

I see what you mean now about the SSList. You shouldn't be able to scroll the green away. I was more focused on how the green line looked on the left of the scrollbar (vs the right). Would it make sense to open issues? I think the SSList/Scroll pane one can probably be pretty well described. It occurred to me that  there could be a similar issue with SSTextArea if a scroll pane is added. **_(2024-03-20 edit: I confirmed that an SSTextArea is doing the same thing when scroll bars are involved.)_**

Not if you have enough to write up about the 'lurking' image/border issue. 

Much less likely to be forgotten if we open issues.

PR #189, opened 2024-02-20
Formatting - move factories

bpangburn:
Moving factory classes into their own subdirectory. Added UTF-8 encoding based on some Eclipse complaints.

Plan to have a few formatting related branches for which I'll open PRs. When good progress is made we can PR "Formatting" into "4.0.13-SNAPSHOT"


errael (approved):
Scanned through "Files changed". OK. Assuming still compiles.

PR #190, opened 2024-02-20
Deleting deprecated classes.

bpangburn:
Deleting 11 classes that are deprecated at the class (vs method or variable) level.


errael (approved):
Scanned through "Files changed". OK, less is more. Assuming still compiles.

PR #192, opened 2024-03-14
Remove final two switch(sqlType) statements from `SSTableModel`.

errael:
Remove final two switch(sqlType) statements from `SSTableModel`.
Add `getColumnObject()` to `RowSetOps`.

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
@@ -894,4 +894,95 @@ public class RowSetOps {
 		}
 		
 	}
+
+	/**
+	 * Fetch and return object for specified rowset's specified column
+	 * from the database; object type is determined by column's JDBCType.
+	 * There is no filtering, for example null conversion.
+	 * @see <a href="https://download.oracle.com/otn-pub/jcp/jdbc-4_3-mrel3-eval-spec/jdbc4.3-fr-spec.pdf">JDBC 4.3 Specification</a> Appendix B-1
+	 * @param rowset source rowset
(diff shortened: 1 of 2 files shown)


swingset/src/main/java/com/nqadmin/swingset/datasources/RowSetOps.java
```diff
@@ -894,4 +894,72 @@ public static void updateColumnObject2(RowSet _rowSet, Object _value, int _colum
...
+	 * Fetch and return object for specified rowset's specified column;
+	 * type is determined by column's JDBCType.
+	 * @param rowset source rowset
+	 * @param _column source column
+	 * @return 
+	 * @throws SQLException
+	 */
+	public static Object getColumnObject(RowSet rowset, int _column) throws SQLException
+	{
+		if (Boolean.TRUE)
```
bpangburn: Won't `if (Boolean.TRUE)` always resolve to true?

errael: Right. I think it's OK to replace the 40+ lines by a single line. But that's based on my experience with `H2`; looking at specs and h2 code. I'm not that familiar with other JDBC drivers. I have learned through reading that they can vary greatly. In fact `SQLFeatureNotSupportedException` is listed for almost every method in ResultSet javadoc. So this change leaves both the original and one liner as something that can be easily switched.

errael:
Changes in comments: see https://github.com/bpangburn/swingset/compare/65855b120975e109aea8211adcc2c5dc80b3cb2d..5d3ab2b4611e76ebf50d61ffc98921a63ee599fc .

bpangburn (approved):
Looks good.

PR #193, opened 2024-03-16
Decorate

bpangburn:
Draft PR to consider merging Decorate into formattting-WORKING rather than 4.0.13-SNAPSHOT.


bpangburn:
Appears to be one less conflict merging into formatting-WORKING vs 4.0.13-SNAPSHOT, but that conflict might reappear merging formatting-WORKING into 4.0.13-SNAPSHOT.

bpangburn:
Since Decorate is general purpose and impacts components outside of SSFormattedTextField we'll go with original PR#186

PR #194, opened 2024-03-20
Grid colored multi key sort arrows

errael:
Extend sort infrastructure to handle multi-key arrows.
Color shows sort order, configure colors in GridTableCellHeaderRenderer.
Could use gray colors.

multi-key-sort

### DIFF
--- a/swingset/src/main/java/com/nqadmin/swingset/SSDataGrid.java
+++ b/swingset/src/main/java/com/nqadmin/swingset/SSDataGrid.java
@@ -52,10 +52,8 @@ import java.sql.JDBCType;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Calendar;
 import java.util.Collections;
 import java.util.EventObject;
-import java.util.GregorianCalendar;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
@@ -81,6 +79,7 @@ import javax.swing.SortOrder;
 import javax.swing.SwingConstants;
 import javax.swing.border.LineBorder;
 import javax.swing.table.DefaultTableCellRenderer;
+import javax.swing.table.JTableHeader;
 import javax.swing.table.TableCellRenderer;
 import javax.swing.table.TableColumn;
 import javax.swing.table.TableColumnModel;
@@ -90,6 +89,7 @@ import javax.swing.table.TableRowSorter;
 import org.apache.logging.log4j.Logger;
 
 import com.nqadmin.swingset.datasources.RowSetOps;
+import com.nqadmin.swingset.models.GridTableHeader;
 import com.nqadmin.swingset.models.SimpleComboListSwingModels;
 import com.nqadmin.swingset.utils.SSCommon;
 import com.nqadmin.swingset.utils.SSUtils;
@@ -833,6 +833,12 @@ public class SSDataGrid extends JTable {
 	@Override
 	protected SSTableModel createDefaultDataModel() {
 		return new SSTableModel();
+	}
+
+	/** {@inheritDoc } */
+	@Override
+	protected JTableHeader createDefaultTableHeader() {
+		return new GridTableHeader(columnModel);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


bpangburn:
Nice addition and seems like a no-brainer to add to grid-WORKING. My concern is that without user documentation somewhere (or experimentation), users might not understand the sort order from the colors.

What do you think about having a number (maybe in parenthesis)  to the right of the up/down arrow. Given columns Xray, Yankee, Zulu and sort (ASCending) of Y, Z, X then you'd have something like:
Xray ^ (3) | Yankee ^ (1) | Zulu ^ (2)

errael:
Some missing information. @prasanthreddy-git and I have been discussion this in email. This is just one example. The main idea here is an example of how this can be done as part of the sorting infrastructure, rather than as part of the app.

> Nice addition and seems like a no-brainer to add to grid-WORKING. My concern is that without user documentation somewhere (or experimentation), users might not understand the sort order from the colors.

If the user knows enough to do a multi-key sort, I'd think they can notice the heading as they are clicking. But doc is important, two things that come to mind:
- Tool tip with description. Maybe the header of a column that's part of the sort, has a tooltip with it's position in the sort.
- Use gray scale, rather than colors. Black is primary key, dark gray is secondary key, light gray is tertiary key.

> What do you think about having a number (maybe in parenthesis) to the right of the up/down arrow. Given columns Xray, Yankee, Zulu and sort (ASCending) of Y, Z, X then you'd have something like: Xray ^ (3) | Yankee ^ (1) | Zulu ^ (2)

bpangburn:
Grayscale is probably the easiest, but less exciting than colors and tooltips. Up to you or I can accept the PR as-is and give @prasanthreddy-git  a jar to test all of grid-WORKING with our platform.

errael:
I'm such a sucker for a challenge; the colors do add a bit of spice. And of course can't just slam in a toolTip, must have a way for the developer to provide their own information...

The capture program left out the cursor which is on the green arrow

sortToolTip

errael:
I was concerned that the light gray arrow would blend into the background. (yeah, right)

bpangburn:
@errael I was about to correct your spelling of 'tertiary' with 'ternary', but then I would have looked like a jackass as tertiary is the correct term.

I do have a suggestion and @prasanthreddy-git might want to weigh in too. I think the sorting should be reversed. Suppose I have  3 columns, first_name, middle_name, & last_name, and I want to sort them last, first, middle. I would expect to click last_name, first_name, middle_name, but it seems you're doing the opposite. If I'm using a database with a UI like Microsoft Access, I would open the query designer, select the table, select the fields I need, and then similar to the above, I'd select Ascending for last, first, and then middle. If I'm writing an ORDER BY clause in SQL, I'd type, `ORDER BY last_name, first_name, middle_name`. ASC is normally the default so I left that off.

errael:
> @errael I was about to correct your spelling of 'tertiary' with 'ternary', but then I would have looked like a jackass as tertiary is the correct term.

I went through exactly that earlier today.

> I do have a suggestion and @prasanthreddy-git might want to weigh in too. I think the sorting should be reversed. Suppose I have 3 columns, first_name, middle_name, & last_name, and I want to sort them last, first, middle. I would expect to click last_name, first_name, middle_name, 

I don't disagree.

> but it seems you're doing the opposite. 

Key manipulation after a click, like inserting the new key in the comparison function, is handled by Swing. Not to say it could be worked around, but it's not so straightforward.

> If I'm using a database with a UI like Microsoft Access, I would open the query designer, select the table, select the fields I need, and then similar to the above, I'd select Ascending for last, first, and then middle. If I'm writing an ORDER BY clause in SQL, I'd type, `ORDER BY last_name, first_name, middle_name`. ASC is normally the default so I left that off.

Is it used by people who write SQL? :-)

bpangburn:
> Key manipulation after a click, like inserting the new key in the comparison function, is handled by Swing. Not to say it could be worked around, but it's not so straightforward.

That's unfortunate. Was hoping it would be an easy change. I can certainly live with it like it is.

errael:
> first_name, middle_name, & last_name, and I want to sort them last, first, middle. I would expect to click last_name, first_name, middle_name

That reminds me of a quote from a movie, might have been Brazil, when asking for a form to be filled out. Said quickly and bruskly: "last name first first name middle name last"

PR #195, opened 2024-03-21
Adding sort column number to the header along with up/down arrows.

prasanthreddy-git:
Adding sort order number and icons.


errael:
Since this code is doing something in the Application that I believe should be done in the infrastructure, I'll decline to review it.

PR #205, opened 2024-11-29
New .formatting classes

bpangburn:
My naming is on this is backwards. The mashup of SwingSet 4.1.0-SNAPSHOT and @errael SS-5 is in 4.1.0-SNAPSHOT. 4.1.0-MASHUP is basically identical to RC2 (e.g. prior to bringing all of SS-5 .formatting into 4.1.X).

Lot of changes. Looking for sanity checks in case I did something stupid.

Tried to mark places where old code is retained over newer code with a TODO:
`TODO: 2024-11-29_BP: Disabling this for 4.1.x. Add back for 5.x.`

Primary objective here was to bring in all of the work done on .formatting without bringing in many other changes/features and avoid introducing new jar dependencies. Will look at adding dependencies once we have a good 4.1.x released.


errael:
About the partialSSCommon stuff. There was the old constructingSSCommon which
was used in SSTextField::createDefaultModel. I ran into another case, probably
in SSFTF, where there was an issue. So I generalized it and put it everywhere.
At some point it would be nice to get rid of this, but that requires a new
component initialization strategy.

I'm not sure how much stuff brought in is actually used. Thought I'd run it
first to see how it goes before looking at files.

Using command line
- I pulled, applied PR #205 to MASHUP. To build: `mvn clean package -Prelease`
- Build swingset (disabled javadoc).
```
    [INFO] -------------------< com.nqadmin.swingset:swingset >--------------------
    [INFO] Building swingset 4.1.0-SNAPSHOT
```
- Build of swingset-demo failed.
```
    [INFO] --------------< com.nqadmin.swingset.demo:swingset-demo >---------------
    [INFO] Building swingset-demo 4.1.0-SNAPSHOT
```
I also tried with 4.1.0-SNAPSHOT, got something similar.

LMK when it builds, or if it looks like a cockpit error, and I'll try again.

I noticed the latest TestFormattedComponents is not there.
It's mostly allowNull and DebugField, but maybe that has something to do
with it.

bpangburn:
@errael I'm not sure why `mvn clean package -Prelease` wasn't working for you. I'm running the same command against 4.1.0-SNAPSHOT (which has all of the changes, Java 21 code, etc), and other than the JavaDoc, it everything compiled.

I guess confirm that you're running Java 21 and that the $JAVA_HOME variable is properly set?

I don't know why the JavaDoc is exploding in an ugly fashion. Guessing something with newer JDKs. I'll look into it more.

Regarding ssCommon, I saw that you added code to deal with the constructor-issue throughout and I brought all of that code over.  I agree it should be private and there should not be a setter. I don't recall if it was the .formatting classes and new masks that brought the incomplete construction of ssCommon to light or something else. I think I mentioned in an email that I tried using a Initializer Block  to initialize ssCommon and that didn't help. I agree that it would be nice to come up with another solution that didn't require so much code duplication.

errael:
> brought the incomplete construction of ssCommon to light or something else

It first came in with #178 to fix #175.

bpangburn:
Want to take another pass at reconciling 4.1.0-SNAPSHOT and /errael/SS-5 now that we have the local branch SS-5-FROM-ERRAEL

